PR #8, opened 2022-09-13
farm-ng-core protos

strasdat:


CMakeLists.txt
```diff
@@ -67,16 +67,25 @@ set(CMAKE_CXX_FLAGS "")
 set(EXPECTED_BUILD_TESTS FALSE)
 
 find_package(fmt REQUIRED)
+find_package(Protobuf)
```
ethanrublee, line 70: You could mark this as QUIET if you like to keep it optional.

farm_ng/core/CMakeLists.txt
```diff
@@ -5,3 +5,6 @@ core
 add_subdirectory(logging)
 add_subdirectory(enum)
 add_subdirectory(misc)
+if(Protobuf_FOUND)
+  add_subdirectory(serial)
```
ethanrublee, line 9: Suggest a different name than serial (as it might conflict with embedded serial UART terminology).  Perhaps proto, protobuf, or serialization.

farm_ng/core/serial/proto_reader_writer.h
```diff
@@ -0,0 +1,69 @@
...
+/// TODO: Add API docs.
+
+void writeProtobufToJsonFile(
+    const std::filesystem::path& path, const google::protobuf::Message& proto);
+
+void writeProtobufToBinaryFile(
+    const std::filesystem::path& path, const google::protobuf::Message& proto);
+
+template <class ProtobufT>
+ProtobufT readProtobufFromJsonFile(const std::filesystem::path& path) {
```
ethanrublee, line 28: is it possible to hid the implementation of this as much as possible to avoid compiler overhead? Suggest calling a non-templated function from the templated function, that takes a ``google::protobuf::Message* out_proto`` output parameter.

PR #9, opened 2022-09-19
introduce top-level cpp-dir

strasdat:
- introduction of cpp sub-dir
 - also slightly lower requirement on CMake 3.23 instead of 3.24


.github/workflows/ci.yml
```diff
@@ -84,12 +84,15 @@ jobs:
...
-      run: ./run_clang_tidy.sh
+      run: |
+        cd cpp
+        ./run_clang_tidy.sh
       if: matrix.os == 'ubuntu-20.04'
 
 
     - name: Configure And Build Tests
       run: |
+        cd cpp
```
isherman, line 95: Can also be `working-directory: cpp` at the step level

ethanrublee:
Thank you, looks great.

PR #10, opened 2022-09-20
Initial proto events protos and python classes

edgarriba:
TODO:

- [ ] define `EventPayload` as discussed offline that contains `Any` 
```bash
Event
EventPayload - has an Any
Event
EventPayload - has an Any
```
- [ ] fix paths and package namespace in protos [python]


protos/event.proto
```diff
@@ -0,0 +1,29 @@
+syntax = "proto3";
```
strasdat, line 1: This looks all non-controversial. 
One question is how to structure protos. All in one file for core?

What is a little tricky - getting include path right on the c++ side.
I suggest to merge https://github.com/farm-ng/farm-ng-core/pull/8 first and then place this on top of it.

protos/event.proto
```diff
@@ -0,0 +1,29 @@
...
+enum UriSchemeType {
+  UNKNOWN = 0;
+  FILE = 1;
+  DATA = 2;
+  DEVICE = 3;
+  OTHER = 4;
+}
+
+message Uri {
+  UriSchemeType scheme = 1;
```
strasdat, line 14: In general, I prefer string for enum, since it is easier to parse on the c++ side using FARM_ENUM. (Unless size matters, e.g. if we have a tensor of enums...)

But I'm okay to use ints for enums if that makes it easier on the python side - in general.

But for this case, we should use a string for sure since the scheme of a URL is a string by definition ("A non-empty scheme component followed by a colon (:), consisting of a sequence of characters beginning with a letter and followed by any combination of letters, digits").

strasdat: Also a string makes the json representation of the proto human readable and I believe the URI is part of the header and hence should be human readable.

protos/event.proto
```diff
@@ -0,0 +1,29 @@
...
+}
+
+message Event {
+  string module = 1;  // the class module name
+  string name = 2;    // the class name
+  int64 length_next = 3;
+  Uri uri = 4;
+}
+
+message EventHeader {
```
ethanrublee, line 27: Let's remove the header.  This seems application specific (robot_name); and it could just be by convention the first event in the stream.  I don't think we have a mature enough header to commit it yet.

protos/event.proto
```diff
@@ -0,0 +1,29 @@
...
+message Uri {
+  UriSchemeType scheme = 1;
+  string authority = 2;
+  string path = 3;
+  string query = 4;
+}
+
+message Event {
+  string module = 1;  // the class module name
+  string name = 2;    // the class name
```
ethanrublee, line 22: should this be type_name?  In general, i think the event needs to include:

- the typename (and where to load the type from, shared library or python package) - this information could be stored in the log once for any name.  Perhaps the URI is sufficient for the type information.
- a monotonic timestamp (with a reference clock name) which can be used to order the events in a playback tool

strasdat: > a monotonic timestamp (with a reference clock name) which can be used to order the events in a playback tool

It would be good to document what the semantic of such monotonic timestamp is.

  - Is it the point in time the underlying measurement occurred (e.g. photon hit sensor of imager),
  - the time the data asset was processed/produced,
  - the time the event was serialized,
  - or are we explicit that it could mean any of the above and/or something different
  
    - does monotonic imply that events are strictly order by such timestamp (for that given clock)?
  
  I almost would vote for "time the event was serialized": since it is well defined for any type of event, and by definition strictly ordered (for monotonic clock). Any more detailed timing information can be stored inside the event data.

strasdat:
The Event / EventHeader schema is hard to review without having examples here. But if we are okay with an iterative design approach, we can merge these protos more or less as is, and then modify as needed once we add protos to sophus and build corresponding playback tools.

edgarriba:
Sure, I’m working on that during the weekend

protos/event.proto
ethanrublee: when you say the "time the event was serialized" do you mean when the payload is being written to the the log file?

I think at a generic level we are interested in the timing of all the above; could we have a list of timetstamps, where the struct was:

```
message EventStamp {
double seconds = 1;
Uri clock = 2; // encodes semantic of when stamp was sampled, and from which clock domain
double certainty = 3; // what is the 3 sigma error in the stamp, e.g. if the stamp is an estimate.
};

message Event {
   repeated EventStamp stamps = 1; // an event can have multiple stamps; e.g. when it was written to the event log file, when the data was received over grpc, when the camera was triggered, when the camera began and ended exposure.
   Uri uri = 2; // URI could encode type name (which is used for look up for where to load the event from).
   int64 length_next = 3;
};
```

I advocate for the above because timing is a fundamental aspect of distributed systems and control/estimation, and it would be great to have access to this timing rich timing information for understanding a system from the logs in a generic way.

strasdat: Yeah, the above works. I like the design.

I'd like to understand the Event message a ittle better and pros and cons storing large binary blobs inside the stream vs. as a side cart.

PR #16, opened 2022-09-27
fixed python code

edgarriba:


py/farm_ng/core/uri.py
```diff
@@ -13,39 +17,26 @@ def __repr__(self) -> str:
...
-        self, scheme_name: str, authority: str, path: str, query: str
-    ) -> "Uri":
-        scheme_value = event_pb2.UriSchemeType.DESCRIPTOR.values_by_name[
-            scheme_name.upper()
-        ]
-        return Uri(
-            event_pb2.Uri(
-                scheme=scheme_value.number, authority=authority, path=path, query=query
-            )
+    def from_message(cls, message: Any) -> "Uri":
```
ethanrublee, line 38: I think we should enforce that message is a protobuf message, and get the type information from the protobuf, not python module, so that it's the same in c++ and python

ethanrublee (commented):
Wip review

py/farm_ng/core/events_file_reader.py
```diff
@@ -13,17 +13,13 @@
 
 
 class EventsFileReader:
-    def __init__(self, file_name: Path) -> None:
-        self._file_name = file_name.absolute()
-        assert Path(self._file_name.parents[0]).is_dir()
-
+    def __init__(self) -> None:
         self._file_stream: Optional[BinaryIO] = None
```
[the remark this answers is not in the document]
edgarriba: yes, i'll update later -- this is code coming from before we added asyncio in the writer. Reader i believe we don't implement as async.

py/farm_ng/core/uri.py
edgarriba: sure, open to discuss implementation details for this

py/farm_ng/core/events_file_writer.py
```diff
@@ -31,25 +29,22 @@ def is_closed(self) -> bool:
...
+        # TODO: add timestamps
         event = event_pb2.Event(
-            module=str(message.__module__),
-            name=str(message.__class__.__name__),
-            length_next=message.ByteSize(),
             uri=uri,
+            payload_length=message.ByteSize(),
         )
 
         event_len: bytes = event.ByteSize().to_bytes(1, sys.byteorder)
```
ethanrublee, line 50: I don't think this is a safe assumption that your number of bytes is less than 255 for an Event.

edgarriba: 4byte maybe?

PR #19, opened 2022-10-05
Import reader/writer from amiga-brain-api

edgarriba:
replacement of https://github.com/farm-ng/farm-ng-core/pull/16

- imports reader/writer code from https://github.com/farm-ng/amiga-dev-kit/tree/main/brain/farm_ng/core so that we can invert import chain
- slightly change the reader/writer to adjust to the new event file definition

**TODO**

- [ ] pyliny/mypy don't recognise well the generated proto stubs

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,2 +1,5 @@
+[MASTER]
+load-plugins=pylint_protobuf
+
 [MESSAGES CONTROL]
 disable=missing-module-docstring,missing-class-docstring,missing-function-docstring
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,9 @@
+{
+    "python.testing.pytestArgs": [
+        "py"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true,
+    "python.linting.enabled": true,
+	"python.linting.pylintEnabled": true
+}
--- a/py/README.md
+++ b/py/README.md
@@ -1,0 +1,46 @@
+# farm-ng-core [python]
+
+## Install
+
+Install `pip3` & `virtualenv`:
+
+```bash
+sudo apt-get install python3-pip
+sudo pip3 install virtualenv
+```
+
+Clone the project:
+
+```bash
+git clone https://github.com/farm-ng/farm-ng-core.git
+```
+
+Move to the python code directory:
+
+```bash
+cd farm-ng-core/py
+```
+
+Start a virtual environment:
+
+```bash
+# assuming you're already in the amiga-brain-api/ directory
+python3 -m venv venv
+source venv/bin/activate
+```
+
+Create and install the `farm_ng.core` package:
+
+```bash
+# install to system
+pip3 install .
+
+# or for development mode
+pip3 install -e .[dev]
+```
+
+Verify that you have installed the package
+
+```bash
+python3 -c "import farm_ng; print(farm_ng.__version__)"
+```
--- a/py/README.md
+++ b/py/README.md
@@ -24,7 +24,7 @@ cd farm-ng-core/py
 Start a virtual environment:
 
 ```bash
-# assuming you're already in the amiga-brain-api/ directory
+# assuming you're already in farm-ng-core/py directory
 python3 -m venv venv
 source venv/bin/activate
 ```
--- a/py/farm_ng/core/events_file_reader.py
+++ b/py/farm_ng/core/events_file_reader.py
@@ -1,4 +1,5 @@
 import importlib
+from importlib.resources import Package
 import sys
 from collections import defaultdict
 from pathlib import Path
@@ -13,6 +14,16 @@ import logging
 from farm_ng.core import event_pb2
 from farm_ng.core.uri import Uri
 from farm_ng.core import uri_pb2
+
+
+def parse_protobuf_descriptor(desc) -> Tuple[str, str]:
+    # parse a proto descriptor and extract the message name and package.
+    # See: https://developers.google.com/protocol-buffers/docs/reference/python-generated#invocation
+    # NOTE: the descriptor comes in the shape of `farm_ng.core.proto.Timestamp`
+    desc_list = desc.split(".")
+    name = desc_list[-1]
+    package = desc_list[:-2] + [name.lower() + "_pb2"]
+    return name, ".".join(package)
 
 
 class EventsFileReader:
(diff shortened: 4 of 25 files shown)


py/farm_ng/core/events_file_reader.py
```diff
@@ -111,43 +127,56 @@ def num_frames(self, uri: Uri) -> int:
...
-        event: bytes = self._file_stream.read(event_len)
+        event: bytes = file_stream.read(event_len)
 
         event_proto = event_pb2.Event()
         event_proto.ParseFromString(event)
 
         # parse the message
         message_cls = getattr(
-            importlib.import_module(event_proto.module), event_proto.name
+            importlib.import_module(event_proto.uri.scheme), event_proto.uri.authority
```
ethanrublee, line 160: I think we need to document and check how encode this in the uri. Ideally the protobuf type name (there is a descriptor in the protobuf) is used in the uri, and then this is used to look up the python package to import. In c++, a different mechanism will be used to automatically dlopen the shared library that has the protobuf definitions and then look up the c++ symbol name.

ethanrublee (changes requested):

py/README.md
```diff
@@ -0,0 +1,46 @@
...
+Move to the python code directory:
+
+```bash
+cd farm-ng-core/py
+```
+
+Start a virtual environment:
+
+```bash
+# assuming you're already in the amiga-brain-api/ directory
```
ethanrublee, line 27: I don't think Amiga-brain is relevant here.

py/farm_ng/core/events_file_writer.py
```diff
@@ -36,24 +41,32 @@ def open(self) -> bool:
...
+        maybe_file_stream = self._file_stream
+        if maybe_file_stream is None:
+            return
+        file_stream = cast(IO, maybe_file_stream)
 
-        if uri is None:
-            uri = event_pb2.Uri()
+        # create a Uri to store the descriptor of the message
+        # it comes in the form of `farm_ng.core.proto.Timestamp`
+        uri = uri_pb2.Uri(scheme=message.DESCRIPTOR.full_name)
```
edgarriba, line 61: @ethanrublee

py/farm_ng/core/events_file_reader.py
```diff
@@ -1,83 +1,97 @@
...
+
+
+def parse_protobuf_descriptor(desc) -> Tuple[str, str]:
+    # parse a proto descriptor and extract the message name and package.
+    # See: https://developers.google.com/protocol-buffers/docs/reference/python-generated#invocation
+    # NOTE: the descriptor comes in the shape of `farm_ng.core.proto.Timestamp`
+    desc_list = desc.split(".")
+    name = desc_list[-1]
+    package = desc_list[:-2] + [name.lower() + "_pb2"]
+    return name, ".".join(package)
```
edgarriba, line 26: @ethanrublee this reverse engineer how protobuf compiles

py/farm_ng/core/events_file_reader.py
```diff
@@ -104,52 +119,46 @@ def close(self) -> bool:
...
 
         event_proto = event_pb2.Event()
         event_proto.ParseFromString(event)
 
         # parse the message
-        message_cls = getattr(
-            importlib.import_module(event_proto.module), event_proto.name
-        )
+        name, package = parse_protobuf_descriptor(event_proto.uri.scheme)
+        message_cls = getattr(importlib.import_module(package), name)
```
edgarriba, line 157: @ethanrublee not sure if this what you meant at all. I checked the descriptors api but i couldn't find a way to dynamically create descriptors/classes using the native protobuf api.

.pylintrc
```diff
@@ -1,2 +1,5 @@
+[MASTER]
+load-plugins=pylint_protobuf
```
ethanrublee, line 2: @edgarriba this still doesn't work for me.  https://github.com/nelfin/pylint-protobuf/issues/19

py/farm_ng/core/events_file_reader.py
ethanrublee: yes this was what I was thinking.

.pylintrc
ethanrublee: Looks like it may be caused by a recent version of protobuf: https://github.com/protocolbuffers/protobuf/issues/10372

ethanrublee:
@strasdat @edgarriba this is just about ready to go in.  Currently i don't have a test that writes from c++ and then reads from python, or vice versa, but I think that can be added later.

The format is relatively simple, and i think we should have a few small follow on pull requests to add features required as needed, from the amiga-dev-kit side.

Please note, on the c++ side, I haven't made a look up from the uri.query type string to corresponding c++ protobuf.

py/farm_ng/core/events_file_reader.py
```diff
@@ -2,46 +2,59 @@
...
+            return []
         return [Uri.from_string(key) for key in self.offsets.keys()]
 
     def has_uri(self, uri) -> bool:
+        if self.offsets is None:
+            return False
         return uri.string() in self.offsets.keys()
 
-    # TODO: discuss the api signature
-    def compute_offsets(self) -> Dict[Uri, List[int]]:
```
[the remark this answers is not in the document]
ethanrublee: tried to fix. now building the index is lazy.

py/farm_ng/core/events_file_writer.py
ethanrublee: I made this part of the query, and used scheme="protobuf", query="type=farm_ng.core.proto.Timestamp",

py/farm_ng/core/events_file_reader.py
ethanrublee: note i found this does not work for protobufs in general, like from google.  This makes an assumption about how the protobuf packages are organized, so only works for farm-ng protobufs, which is OK for now.

ethanrublee (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
+    "python.testing.pytestArgs": [
+        "py"
+    ],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true,
+    "python.linting.enabled": true,
+	"python.linting.pylintEnabled": true
```
edgarriba, line 8: @ethanrublee adding this seems to work fine in vscode but still not recognized during the pre-commit stage

py/farm_ng/core/events_file_reader.py
edgarriba: @ethanrublee this definitely won't work after pulling this to the adk. E.g an `OakFrame` gets the following package name `ModuleNotFoundError: No module named 'farm_ng.oak.oakframe_pb2'

ethanrublee: Yes we will need to port that data to this new format.

.vscode/settings.json
hothanhloc68: "pm Grant by4a.setedit22 android.permission.WRITE_SECURE_SETTINGS".

PR #21, opened 2022-10-10
fix CI - POSIX related issue

strasdat:


cpp/farm_ng/core/prototools/event_log_writer.cpp
```diff
@@ -9,6 +9,8 @@
 #include "farm_ng/core/prototools/event_log_writer.h"
 
 #include <farm_ng/core/logging/logger.h>
+
+#define __USE_POSIX
```
ethanrublee, line 13: oh thanks, sorry I didn't catch this.

PR #27, opened 2022-10-15
Zipping working well, breaking api change.

ethanrublee:


cpp/farm_ng/core/prototools/event_log_reader.cpp
```diff
@@ -158,7 +158,9 @@ bool EventTimeCompare::operator()(
   auto maybe_lhs_stamp = getStamp(lhs.event(), clock_name, semantics);
   auto maybe_rhs_stamp = getStamp(rhs.event(), clock_name, semantics);
   if (maybe_lhs_stamp && maybe_rhs_stamp) {
-    return maybe_lhs_stamp->stamp() < maybe_rhs_stamp->stamp();
+    return maybe_lhs_stamp->stamp() < maybe_rhs_stamp->stamp() ||
+           (!(maybe_rhs_stamp->stamp() < maybe_lhs_stamp->stamp()) &&
+            lhs.event().uri().path() < rhs.event().uri().path());
   }
   return false;
```
strasdat, line 165: @ethanrublee  Just noticed the false here.

This means that the operator does not satisfy **Strict Weak Ordering** because it violates:
"f(x, y) implies !f(y, x)" https://www.boost.org/sgi/stl/StrictWeakOrdering.html

Using this comparison operator in functions like std::sort: https://medium.com/@shiansu/strict-weak-ordering-and-the-c-stl-f7dcfa4d4e07

Since UB is strictly worse than a crash I would suggest either:

1. To crash when the stamp is not available or similar. 
2. Rename the function "fooBar()" and document that it is not strictly weak ordered.

strasdat (changes requested):
See my comment about potential UB. Continuing reviewing the rest of the code.

cpp/farm_ng/core/prototools/event_log_reader.cpp
```diff
@@ -20,54 +20,67 @@ namespace farm_ng {
...
+    in = std::ifstream(log_path.string(), std::ofstream::binary);
     if (!in) {
-      throw std::runtime_error(
-          FARM_FORMAT("Could not open file: {}", log_path));
+      throw EventLogExist(FARM_FORMAT("Could not open file: {}", log_path));
     }
   }
-
   std::string readBytes(uint64_t n_bytes) {
+    FARM_CHECK(in.good());
```
strasdat, line 34: Is this violation a precondition or class invariant. Or should we throw here?

cpp/farm_ng/core/prototools/event_log_reader.cpp
```diff
@@ -20,54 +20,67 @@ namespace farm_ng {
...
     }
     std::streampos pos = in.tellg();
     if (payload) {
       *payload = readBytes(event.payload_length());
     } else {
       in.seekg(event.payload_length(), in.cur);
       if (!in) {
-        throw std::runtime_error("Could not seek past payload.");
+        reset();
+        throw EventLogEOF("Could not seek past payload.");
```
strasdat, line 70: EventLogEOF -> EventLogEof

cpp/farm_ng/core/prototools/event_log_reader.cpp
```diff
@@ -78,34 +91,103 @@ class EventLogReaderBinaryImpl : public EventLogReaderImpl {
...
 }
-}  // namespace
 
-EventLogReader::EventLogReader(std::filesystem::path const& log_path) {
-  resetImpl(impl_, log_path);
+std::vector<EventLogPos> const& EventLogReaderImpl::getIndex() {
+  if (index_.empty()) {
+    reset();
+    while (true) {
+      try {
```
strasdat, line 112: nit: suggest using expected instead.

cpp/farm_ng/core/prototools/event_log_reader.h
```diff
@@ -62,7 +91,26 @@ class EventLogReader {
   void reset();
 
  private:
-  std::unique_ptr<EventLogReaderImpl> impl_;
+  std::shared_ptr<EventLogReaderImpl> impl_;
 };
 
+core::proto::Timestamp const* getStamp(
```
strasdat, line 97: returns pointer, a line of comment what the semantic is (e.g. no memory allocation, n/a stamp -> nullptr)

cpp/farm_ng/core/prototools/event_log_writer_test.cpp
```diff
@@ -44,15 +44,140 @@ TEST(event_log, roundtrip) {
...
+    }
+  }
+  {
+    EventLogReader reader1(file1);
+    EventLogReader reader2(file2);
+    // Grab a reference write stamp from the first reader.
+    core::proto::Timestamp ref =
+        reader1.readNextEvent().event().timestamps()[0];
+    EXPECT_EQ(ref.semantics(), "log/write");
+    std::vector<EventLogPos> index = eventLogTimeOrderedIndex(
```
strasdat, line 111: This can throw, we need a catch block to document to uses canonical use of API.

(Likely there are more throwing functions below.)

strasdat (commented):
Looks great! (modulo issue with ordering)

cpp/farm_ng/core/prototools/event_log_writer_test.cpp
```diff
@@ -44,15 +44,140 @@ TEST(event_log, roundtrip) {
     EventLogReader reader(file);
     for (int i = 0; i < 10; ++i) {
       std::string payload;
-      auto [event, pos] = reader.readNextEvent(&payload);
+      EventLogPos pos = reader.readNextEvent(&payload);
```
strasdat, line 47: +100

cpp/farm_ng/core/prototools/event_log_reader.cpp
ethanrublee: This a precondition. As the exception should have been handled before this point. Also if an exception is thrown, the io stream is reset.

cpp/farm_ng/core/prototools/event_log_reader.cpp
ethanrublee: Yes this should give an error.

cpp/farm_ng/core/prototools/event_log_writer_test.cpp
ethanrublee: this won't throw unless the file doesn't exist (which will throw in on construction of the reader.

cpp/farm_ng/core/prototools/event_log_writer_test.cpp
```diff
@@ -44,15 +44,140 @@ TEST(event_log, roundtrip) {
...
+    core::proto::Timestamp x;
+    for (int i = 0; i < 10; ++i) {
+      x.set_stamp(i);
+      writer.write("my_stamps", x);
+    }
+  }
+  {
+    EventLogReader reader(file);
+    int i = 0;
+    for (EventLogPos const& pos : reader.getIndex()) {
```
ethanrublee, line 75: @strasdat this api does not throw on file reading,  as it seeks and builds up an index first.  I'll add an example with no index which does throw.

cpp/farm_ng/core/prototools/event_log_reader.cpp
ethanrublee: ok, for now I'll go with exceptions based interface for this library.  There are a few places in the stack that can error out, its a bit easy to code the happy path and let the exceptions propagate.

cpp/farm_ng/core/prototools/event_log_reader.cpp
ethanrublee: Ok make this CHECK fail if this pre-condition is not met that both events contain the stamp and the semantics string.

strasdat (approved):

PR #31, opened 2022-10-19
Fix readJsonStringFromJsonFile method which didn't compile when insta…

stevenlovegrove:
…ntiated.


strasdat:
Thanks for the fix!

PR #35, opened 2022-10-23
Centralize python packaging logic.

ethanrublee:


edgarriba (commented):
LGTM. But what about a `farm_ng_package` in the same philosophy of `farm_ng_cmake`?

ethanrublee:
For now, i think we just make farm-ng-core the basic requirement for python.  We might end up moving farm_ng_cmake back into core not sure its worth the complexity.

ethanrublee:
> LGTM. But what about a `farm_ng_package` in the same philosophy of `farm_ng_cmake`?

On further thought this is a decent idea, so that farm_ng_core doesn't have to bootstrap itself.

ethanrublee:
but for later.

PR #36, opened 2022-10-24
Ensure runtime version of fmt::format is used in C++20 environments f…

stevenlovegrove:
…or this code path.


stevenlovegrove:
libfmt @8.0.1_0 according to macports

PR #38, opened 2022-11-07
Upstream farm-ng-pysetup

edgarriba:
goes with https://github.com/farm-ng/farm-ng-pysetup


setup.py
```diff
@@ -1,19 +1,19 @@
...
-package_py = pathlib.Path(__file__).parent.resolve() / "py/farm_ng/core/"
+# pylint: disable=no-name-in-module, import-error
+from farm_ng.package.package import (
+    build_package_protos,
+    clean_protos,
+)  # pylint: enable=no-name-in-module, import-error
 
 
+# TOOD: move `Command`` classes to farm-ng-pytsetup
 class BuildProtosCommand(Command):
```
ethanrublee, line 17: Now we could move these commands in to farm-ng-package ?

ethanrublee (approved):
Cool

setup.py
edgarriba: Yes, stage 2

PR #39, opened 2022-11-08
update farm-ng-package

edgarriba:


edgarriba:
@ethanrublee i'm not very happy with this pattern but the way it is with setuptools, the `Command` class is very flexible to pass external params programmatically

ethanrublee:
I see. No worries, let's go with it. One question, can you put the option parameters into the setup.cfg and they automagically get set on these commands?

Issue #45, opened 2022-11-18
[Feature] Merge boilerplate `get_event` and `read_message` 


edgarriba:
@Hackerman342 @ethanrublee an improvement here is to merge `get_event` and `read_message` since we use all the time together. For testing and some internal use cases is good but better coupled for a more user friendly api
```python
def read_uri_message(self, uri: uri_pb2.Uri, event_id: int) -> Any:
    # get the event and offset
    event: event_pb2.Event
    offset: int
    event, offset = self.get_event(uri, current_event)

    # seek and parse the message
    return self.read_message(event, offset)
```

maybe `read_message` should be renamed to `parse_message` and `read_message` recive the uri and event_id

_Originally posted by @edgarriba in https://github.com/farm-ng/amiga-brain-api/pull/31#discussion_r1025117076_

ethanrublee:
The c++ side I do like how the event is wrapped in a helper class that has
the read function similar to an itetator, this allows for just in time
reading and performing range based algorithms on the event.

I would suggest we do something similar in python, and also port the time
ordered zipping from c++ in the same feature as it’s a prime use case

PR #47, opened 2022-11-28
Add `EventLogPosition` python wrapper

edgarriba:
implements #45 based on the c++ codebase

Introduces `EventLogPosition` wrapper to make search more flexible


py/farm_ng/core/events_file_reader.py
```diff
@@ -60,9 +68,9 @@ def __exit__(self, type: Any, value: Any, traceback: Any) -> None:
 
     def __repr__(self) -> str:
         return (
-            f"file_name: {str(self.file_name)} "
-            + f"file_stream: {self._file_stream} "
-            + f"is_open: {self.is_open()} "
+            f"file_name: {str(self.file_name)}\n"
+            + f"file_stream: {self._file_stream}\n"
+            + f"is_open: {self.is_open()}"
```
Hackerman342, line 73:
```suggestion
            "\n".join([
                f"file_name: {str(self.file_name)}",
                f"file_stream: {self._file_stream}",
                f"is_open: {self.is_open()}"
            ])
```

py/farm_ng/core/events_file_reader.py
```diff
@@ -95,19 +107,28 @@ def close(self) -> bool:
...
+        self._event_index = []
 
-    def uris(self) -> List[Uri]:
-        if not self._event_index:
+    def get_uris(self) -> List[Uri]:
+        if len(self._event_index) == 0:
             self._build_event_index()
-        return [string_to_uri(key) for key in sorted(self._event_index.keys())]
+
+        uris_set: Set[EventLogPosition] = set()
```
Hackerman342, line 116:
```suggestion
        uris_set: Set[str] = set()
```

py/farm_ng/core/events_file_reader.py
```diff
@@ -95,19 +107,28 @@ def close(self) -> bool:
...
+        for event_log in self.event_index:
+            uris_set.add(uri_to_string(event_log.event.uri))
+
+        return [string_to_uri(x) for x in sorted(uris_set)]
```
Hackerman342, line 120: Can you not make a set of uri's to begin with?

py/farm_ng/core/events_file_reader.py
```diff
@@ -95,19 +107,28 @@ def close(self) -> bool:
...
 
     def has_uri(self, uri: Uri) -> bool:
-        if self._event_index is None:
+        if len(self._event_index) == 0:
             return False
```
Hackerman342, line 124: If using one liner below
```suggestion
```

py/farm_ng/core/events_file_reader.py
```diff
@@ -95,19 +107,28 @@ def close(self) -> bool:
...
-        return uri_to_string(uri) in self._event_index.keys()
 
-    def _read_next_event(self) -> Event:
+        for idx in self.event_index:
+            if uri_to_string(uri) == uri_to_string(idx.event.uri):
+                return True
+        return False
```
Hackerman342, line 129: Optionally
```suggestion
        return uri_to_string(uri) in [uri_to_string(idx.event.uri) for idx in self.event_index]
```

Hackerman342: Probably worse for performance if `self.event_index` is long

py/farm_ng/core/events_file_reader.py
```diff
@@ -45,10 +55,8 @@ def __init__(self, file_name: Union[str, Path]) -> None:
         self._file_stream: Optional[IO] = None
         self._file_length: int = 0
 
-        # store the bytes offset in a dictionary where:
-        # - the key is the string representation of the Uri of the message
-        # - the value is a list with tuples of Event and its offsets to messages with that Uri
-        self._event_index: DefaultDict[str, List[Tuple[Event, int]]] = defaultdict(list)
+        # store index of events in a list to seek faster
+        self._event_index: List[EventLogPosition] = []
```
Hackerman342, line 59: Clearer that it is all events, not the index of a single event
```suggestion
        self._events_index: List[EventLogPosition] = []
```

py/farm_ng/core/events_file_reader.py
```diff
@@ -73,6 +81,10 @@ def file_length(self) -> int:
     def file_name(self) -> Path:
         return self._file_name
 
+    @property
+    def event_index(self) -> List[EventLogPosition]:
+        return self._event_index
```
Hackerman342, line 86: Clearer that it is all events, not the index of a single event
```suggestion
    def events_index(self) -> List[EventLogPosition]:
        return self._events_index
```

py/farm_ng/core/events_file_reader.py
```diff
@@ -135,49 +157,46 @@ def _build_event_index(self) -> None:
...
+        return len(self.get_events(uri))
 
-    def get_event(self, uri: Uri, frame_id: int) -> Tuple[Event, int]:
-        assert self.has_uri(uri)
-        assert frame_id < self.num_events(uri)
-        return self.events(uri)[frame_id]
+    def get_index(self) -> List[EventLogPosition]:
+        if len(self._event_index) == 0:
+            self._build_event_index()
+        return self._event_index
```
Hackerman342, line 181:
```suggestion
    def get_events_index(self) -> List[EventLogPosition]:
        if len(self._events_index) == 0:
            self._build_events_index()
        return self._events_index
```

py/farm_ng/core/events_file_reader.py
```diff
@@ -135,49 +157,46 @@ def _build_event_index(self) -> None:
...
         message: Message = message_cls()
         message.ParseFromString(payload)
         return message
 
     def read(self) -> Tuple[Event, Message]:
         assert self.is_open()
-        event = self._read_next_event()
-        return event, self.read_message(event)
+        event_log: EventLogPosition = self.read_next_event()
+        return event_log.event, self.read_message(event_log)
```
Hackerman342, line 199: `event_log` could be confusing for a single event
```suggestion
        event: EventLogPosition = self.read_next_event()
        return event.event, self.read_message(event)
```

Hackerman342 (approved):

ethanrublee (commented):
I would suggest simplifying the api by removing the uri filtering from this code. Uri filtering can easily be done while iterating over the event pod list, in an application. I think until there are clear common use cases it's better to have less functionality in this reader class so that there is less to maintain and it becomes stable faster. Recommend following c++ interface.

py/farm_ng/core/events_file_reader.py
edgarriba: just replicating the previous behavior where we based on a dict with unique uris, in the this new approach using EventLogPosition we build the index a list not dict

py/farm_ng/core/events_file_reader.py
```diff
@@ -95,19 +113,19 @@ def close(self) -> bool:
...
+    def get_uris(self) -> List[Uri]:
+        if len(self._events_index) == 0:
             self._build_event_index()
-        return [string_to_uri(key) for key in sorted(self._event_index.keys())]
 
-    def has_uri(self, uri: Uri) -> bool:
-        if self._event_index is None:
-            return False
-        return uri_to_string(uri) in self._event_index.keys()
+        uris_set: Set[str] = set()
```
ethanrublee, line 122: Ideally this is built once during the index constructor so that it's constant small time to call.

ethanrublee (approved):

py/tests/test_core.py
```diff
@@ -61,10 +61,10 @@ def test_smoke(self, log_file: Path) -> None:
         with EventsFileReader(log_file) as reader:
             assert reader.is_open()
             assert len(reader.events_index) == 0
-            uris = reader.get_uris()
+            uris = sorted([*{x.event.uri.path for x in reader.get_index()}])
```
ethanrublee, line 64: nice one liner!

edgarriba: very ugly actually :smile:

PR #50, opened 2022-12-14
fix possible UB

strasdat:


ethanrublee (changes requested):

cpp/farm_ng/core/logging/format.cpp
```diff
@@ -16,19 +16,14 @@
 
 namespace farm_ng {
 
-std::function<void(std::string const &)> &getLogLineFunction() {
```
ethanrublee, line 19: This is in the CPp file so that it's only defined once in the shared library

ethanrublee (commented):
Did you encounter undefined behavior?

PR #58, opened 2023-01-09
System clock logging and file size rollover in Python

Hackerman342:


ethanrublee (approved):

ethanrublee:
Can we add a simple test for covering this function in ci?

py/tests/test_core.py
```diff
@@ -98,15 +103,15 @@ def test_parse_proto_descriptor(self):
...
+    def test_write_read(self, log_file: Path, reader_log_file: Path) -> None:
         num_events = 10
         with EventsFileWriter(log_file) as writer:
             for i in range(num_events):
                 time_stamp = timestamp_pb2.Timestamp(stamp=i)
                 writer.write(path="hello", message=time_stamp)
                 writer.write(path="world", message=time_stamp)
 
-        with EventsFileReader(log_file) as reader:
+        with EventsFileReader(reader_log_file) as reader:
```
ethanrublee, line 114: can you extend this test to cause it to roll over?

ethanrublee (approved):

py/tests/test_core.py
Hackerman342: https://github.com/farm-ng/farm-ng-core/pull/58/commits/78a0d5953ede67ce99e45338c6136c473a518079

py/farm_ng/core/events_file_writer.py
```diff
@@ -18,15 +18,25 @@
...
+        """
         if isinstance(file_name, str):
             file_name = Path(file_name)
-        self._file_name: Path = file_name.absolute()
-        assert Path(self._file_name.parents[0]).is_dir()
+
+        assert Path(file_name.parents[0]).is_dir()
+
+        self._file_base: Path = (file_name.parents[0] / file_name.stem).absolute()
+        self.extension: str = file_name.suffix
```
edgarriba, line 32: I'd include this as an optional argument and default to ".bin"

In that way it will be more clear that users provide just the session base name and we then create sub files.

py/tests/test_core.py
```diff
@@ -17,16 +17,26 @@ def fixture_log_file(tmpdir) -> Path:
...
+def fixture_log_dir(tmpdir) -> Path:
+    return Path(tmpdir)
+
+
 class TestEventsWriter:
     def test_smoke(self, log_file: Path) -> None:
         with EventsFileWriter(log_file) as writer:
             assert writer.is_open()
-            assert writer.file_name == log_file
+            # assert writer.file_name == log_file
```
edgarriba, line 34: If we provide base name it will be less confusing that user gives base_name but later file_name is a composed string

py/tests/test_core.py
```diff
@@ -136,3 +146,23 @@ def test_write_read(self, log_file: Path) -> None:
                     assert message.stamp == i
 
         assert reader.close()
+
+    def test_write_rollover(self, log_file: Path, log_dir: Path) -> None:
```
edgarriba, line 150: New properties max_files_length and file_idx should be tested too

edgarriba (changes requested):

py/farm_ng/core/events_file_writer.py
Hackerman342: That's how I had it and changed it in https://github.com/farm-ng/farm-ng-core/pull/58/commits/4a0424d37d2ec1fd6163e7e9029e5c624639fa16. Will revert, as I agree.

Hackerman342:
@edgarriba requested changes are addressed

PR #64, opened 2023-02-06
Add json

strasdat:


farm-ng-core.code-workspace
```diff
@@ -0,0 +1,13 @@
+{
```
edgarriba, line 1: can this file be added  somehow only at c++ level ?

strasdat: yeah, I think I want to get rid of the whole super_project folder soon - similarly as I did for sophus2.

PR #65, opened 2023-02-11
Add log level filtering and log header formatting

isherman:
- Adds support for compile-time and run-time log level filtering.
- Adds support for custom header formatting.
- Adds additional log levels. In general the API tries to align with popular log frameworks like [spdlog](https://github.com/gabime/spdlog).

The log output with the default settings should be the same, or very close to the same, as the current behavior.

I'd like to deprecate `FARM_IMPL_LOG_HEADER` (currently Sophus image_types.h calls directly) and possibly `FARM_PRINTLN`.

In the future a third-party framework like spdlog could offer some compelling features like registering a custom sink that sends text logs to both `stderr` and a binary event log.   

cc: @ethanrublee

### DIFF
--- a/cpp/farm_ng/core/logging/logger.h
+++ b/cpp/farm_ng/core/logging/logger.h
@@ -14,6 +14,7 @@
 
 #pragma once
 
+#include "farm_ng/core/enum/enum.h"
 #include "farm_ng/core/logging/format.h"
 
 #include <farm_pp/preprocessor/comma.hpp>
@@ -31,7 +32,7 @@
 
 // Define FARM_LOG_LEVEL to set the compile-time log level
 #ifndef FARM_LOG_LEVEL
-#define FARM_LOG_LEVEL FARM_LEVEL_INFO
+#define FARM_LOG_LEVEL FARM_LEVEL_DEBUG
 #endif
 
 namespace farm_ng {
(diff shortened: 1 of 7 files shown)


cpp/farm_ng/core/logging/logger_test.cpp
```diff
@@ -12,11 +12,105 @@
...
+
+  // Example 2, custom formatting
+  farm_ng::defaultLogger().setHeaderFormat(
+      "[{time:%Y-%m-%d %H:%M:%S}.{time_ms}]");
+  FARM_INFO("bar");
+  expectContains(
+      capture.buffer(),
+      std::regex{R"(\[\d{4}\-\d{2}\-\d{2} \d{2}:\d{2}:\d{2}.\d{3}\]bar)"});
+
+  // Revert
```
isherman, line 109: Instead of this, just use a non-`defaultLogger` for tests

isherman: nevermind, requires more expressive LOG_ macros

cpp/farm_ng/core/logging/logger.h
```diff
@@ -14,74 +14,267 @@
...
+#define FARM_LEVEL_DEBUG 1
+#define FARM_LEVEL_INFO 2
+#define FARM_LEVEL_WARN 3
+#define FARM_LEVEL_ERROR 4
+#define FARM_LEVEL_CRITICAL 5
+#define FARM_LEVEL_OFF 6
+
+// Define FARM_LOG_LEVEL to set the compile-time log level
+#ifndef FARM_LOG_LEVEL
+#define FARM_LOG_LEVEL FARM_LEVEL_DEBUG
```
strasdat, line 35: Does this mean we are defaulting to FARM_LEVEL_DEBUG? Should't we rather default to FARM_LEVEL_WARN, and then enable per compilation unit if we need more?

strasdat (approved):
Looking great, just some questions about what the default should be...

cpp/farm_ng/core/logging/logger.h
isherman: changed default to INFO (as that's what our current behavior was)

isherman: and that's the default used by other libraries like `spdlog`

PR #66, opened 2023-02-14
Remove #define in logger_test

isherman:
This causes preprocessor multiple definition errors if it's included in a parent project that defines FARM_LOG_LEVEL globally.

We can remove, or put inside a `#ifndef`. 
I admit the correct way to do this is to build the same test with multiple build configurations and see all configurations pass.


isherman:
Closing in favor of https://github.com/farm-ng/farm-ng-core/pull/67

PR #67, opened 2023-02-14
feat: log level per computation unit

strasdat:


cpp/farm_ng/core/logging/logger.h
```diff
@@ -88,7 +84,7 @@ class StreamLogger {
   void flush();
 
   std::string header_format_ = "[FARM {text} in {file}:{line}]";
-  LogLevel log_level_ = LogLevel(FARM_LOG_LEVEL);
+  LogLevel log_level_ = LogLevel(LogLevel::info);
```
strasdat, line 87: The default run-time level need to be managed independently from the default compile-time level.

isherman (commented):
What are the consequences for logging from, e.g. templated code in header files?

strasdat:
closing in favor of #67

PR #71, opened 2023-02-17
feat: print stats about noisy logging if enabled

strasdat:


cpp/farm_ng/core/logging/logger.h
```diff
@@ -67,13 +61,42 @@ class StreamLogger {
       std::string const& function,
       std::string const& message,
       T&&... args) {
+    if (log_level_ <= LogLevel::warning &&
+        (log_level == LogLevel::debug || log_level == LogLevel::trace)) {
+      // We are already in compile-time enabled noisy log-level territory. Hence
```
isherman, line 66: Why not just `log_level_ < LogLevel::warning`?

isherman (approved):

PR #80, opened 2023-04-25
fix: TRY macro, do not evalute expression twice

strasdat:
**Problem:**

The current FARM_TRY is broken since it will evaluate the input expression ``expected_val`` twice. This is problematic if ``expected_val`` is a function call with is either costly or has a side-effect.

***~~Minor Problem:~~***

~~FARM_TRY is too generic of a name for this specific macro.~~

**Solution**:

```FARM_TRY(Type, var, expression)```

- evaluate expression once and assign to temporary variable
- give the temporary variable a unique name (based on the name of ``var``) so there won't be any name classes  when there a multiple FARM_TRY in scope.

***Alternatives:***
 - remove this macro
 - figure out a way to have the temporary variable in a local scope. This is tricky since ``Type var`` needs to be in the outer scope, and we do not want to restrict ``Type`` to being default constructable.

### DIFF
--- a/cpp/farm_ng/core/logging/expected.h
+++ b/cpp/farm_ng/core/logging/expected.h
@@ -74,10 +74,10 @@ struct UnwrapImpl<tl::expected<TT, TE>> {
 #define FARM_UNEXPECTED(cstr, ...) \
   ::tl::make_unexpected(FARM_ERROR_REPORT(cstr, ##__VA_ARGS__))
 
-/// Assigns `*expected_val` to `var` of `Type`, but returns error if there is
+/// Assigns `*expression` to `var` of `Type`, but returns error if there is
 /// one.
-#define FARM_TRY(Type, var, expected_val)             \
-  auto maybe##var = (expected_val);                   \
+#define FARM_TRY(Type, var, expression)               \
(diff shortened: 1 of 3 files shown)


isherman (approved):
The name FARM_TRY was based on precedent: 
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0779r0.pdf
https://www.boost.org/doc/libs/develop/libs/outcome/doc/html/reference/macros/try.html
which are in turn based on conventions in Swift and Rust
so I don't see the need to change the name or function signature.
It would also limit the amount changes to existing code to keep the signature the same.

For a unique name you could append instead of prepending.

That being said, if you feel strongly about this I won't stand in your way.

strasdat:
No I do not feel strongly, just weighing the options.

@isherman I agree with the precedence for "TRY_" - as you pointed out. Also, in Rust the "?" operator was called "try" until the operator was introduced (https://doc.rust-lang.org/std/macro.try.html).

However, this c++ macro does more. It tries to unwrap and  if it succeeds it assigns the unwrap result to a new variable of type Type.

strasdat:
Kept the FARM_TRY name. Thanks for the review!

PR #85, opened 2023-05-25
Naming proposal - no file idx # if not using rollover logging

Hackerman342:


Hackerman342:
@ethanrublee @edgarriba 
Just to start a discussion on this. Seems a little off to name the file `2023_05_19_10_35_13_342989_events_aluminum-pineapple.0000.bin` when the writer is set to not rollover.

Also raises the question of should we remove `.0000` if a log that is using rollover stays under the max file size and there is no `.0001`

edgarriba (approved):
LGTM. Had the same though the other day +1

ethanrublee:
I typically prefer no branching in file names as it makes parsing and file name generation easier when writing scripts to enumerate files. 

I would consider this branching not desirable.

ethanrublee (changes requested):
I don't believe this is a good change to the file format. It makes it less consistent based on how the user records the file, which will have an impact on downstream tools.

Recommend documenting the format, and if it needs changing review the documented format before implementing.

Hackerman342:
Sounds good :+1: . I've documented the naming convention on amiga.farm-ng.com in https://github.com/farm-ng/amiga-dev-kit/pull/186

PR #88, opened 2023-06-05
add event_has_message and improve docs

### LINKED ISSUE #87: [Feature] Add an event filter utility by message type
Add a small utility in the Python API that the allows to filter the events index by types 

```python
def event_has_message(event: event_pb2.Event, msg_type: Any) -> bool:
    """Check if event has a message of type msg_type.

    Args:
        event: event_pb2.Event
        msg_type: protobuf message type

    Returns:
        bool: True if event has a message of type msg_type
    """
    return event.uri.query.split("&")[0].split(".")[-1] == msg_type.__name__
```

edgarriba:
fixes #87

### DIFF
--- a/py/farm_ng/core/events_file_reader.py
+++ b/py/farm_ng/core/events_file_reader.py
@@ -1,18 +1,40 @@
+from __future__ import annotations
 import importlib
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


Hackerman342 (approved):
LGTM

PR #89, opened 2023-06-22
Fix log level macro tests and add additional ASSERT_NEAR test

isherman:
The additional ASSERT_NEAR test is currently failing and requires investigation.


isherman:
Discussed with @stevenlovegrove; it sounds like this macro is intended for checking whether two quantities are the same, within magnitudes of floating-point precision, rather than as a general-purpose comparison of absolute difference against some threshold. 

Couple of suggestions:
- Improve documentation. Consider a warning if threshold is too large, to help prevent misuse. Include documentation of how "threshold" should be interpreted in the case where the relative difference is used inside ASSERT_NEAR. 
- Optionally provide a ASSERT_WITHIN() for the general-purpose usage

strasdat:
Great suggestions: 

I'd suggest to review this crate for inspiration https://docs.rs/approx/0.5.1/approx/

PR #91, opened 2023-07-11
fix tests tmp_path and lints

edgarriba:
- use pytest tmp dir: https://docs.pytest.org/en/6.2.x/tmpdir.html
- fix few co-pilot suggesstions


Hackerman342 (approved):
LGTM

PR #93, opened 2023-07-19
Add smoke for event service and client.

ethanrublee:


protos/farm_ng/core/event_service.proto
```diff
@@ -0,0 +1,85 @@
...
+syntax = "proto3";
+
+import "farm_ng/core/uri.proto";
+import "farm_ng/core/timestamp.proto";
+import "farm_ng/core/event.proto";
+
+package farm_ng.core.proto;
+
+service EventService {
+    rpc getServiceState(GetServiceStateRequest) returns (GetServiceStateReply) {}
```
ethanrublee, line 26: I think this api could be removed, im not sure why its important to track the state of the service at this level.  I think grpc connection state can give us everything we need. @edgarriba @lullabee @Hackerman342

edgarriba: it was originally added to keep tracked of the status of the services in the launcher

py/farm_ng/core/event_service.py
```diff
@@ -0,0 +1,445 @@
...
+
+        event = Event(
+            uri=uri,
+            timestamps=timestamps,
+            payload_length=len(payload),
+            sequence=count,
+        )
+        self._send_event_payload(event, payload)
+
+    # TODO: i'd rename this to `publish` to match with ROS
```
ethanrublee, line 294: This isn't really publish, it allows a client to send a message directly to a service, without the service establishing a subscription. Publish is implemented by sending messages to all the subscribers.

We don't have a centralized broker here. The alternative is to remove this function and declare all subscriptions in the config file, but that seems complex and doesn't allow a simple client to come up and start sending messages directly to a service.

Note these messages currently wouldn't be logged as they are point to point, was investigating republishing them.

This specific interface is still wip for sure.

edgarriba: Do we want/need a centralised broker? That might solve the issue to pass through messages over different services

py/farm_ng/core/event_service_recorder.py
```diff
@@ -0,0 +1,109 @@
...
+            while True:
+                event, payload = await self.record_queue.get()
+                event.timestamps.append(
+                    get_monotonic_now(semantics=StampSemantics.FILE_WRITE)
+                )
+                writer.write_event_payload(event, payload)
+
+    async def subscribe(self, client: EventClient, subscription: SubscribeRequest):
+        async for event, payload in client.subscribe(subscription, decode=False):
+            self.record_queue.put_nowait((event, payload))
```
edgarriba, line 59:
```suggestion
            await self.record_queue.put((event, payload))
```

py/farm_ng/core/event_service_tool.py
```diff
@@ -0,0 +1,172 @@
...
+        args.output,
+        EventServiceConfigList(
+            configs=[
+                EventServiceConfig(
+                    name=args.name,
+                    host=args.host,
+                    port=args.port,
+                    subscriptions=[
+                        SubscribeRequest(
+                            uri=Uri(path="/test", query="service_name=bar"), every_n=1
```
edgarriba, line 50: @ethanrublee how do you subscribe e.g to a specific stream of an oak camera? or which granularity we want in the services

py/farm_ng/core/event_service_recorder.py
```diff
@@ -0,0 +1,109 @@
...
+                event.timestamps.append(
+                    get_monotonic_now(semantics=StampSemantics.FILE_WRITE)
+                )
+                writer.write_event_payload(event, payload)
+
+    async def subscribe(self, client: EventClient, subscription: SubscribeRequest):
+        async for event, payload in client.subscribe(subscription, decode=False):
+            await self.record_queue.put((event, payload))
+
+    async def subscribe_and_record(
```
ethanrublee, line 61: Actually we want no await puts, so we don't queue messages, and small queue sizes. We may drop messages but want this to not put back pressure on sender.

ethanrublee (commented):
@edgarriba

lullabee:
Maybe putting a warning if we drop message

py/farm_ng/core/event_service.py
ethanrublee: I would prefer not to implement a broker yet, I think a config file will be sufficient for a while.

ethanrublee: I renamed this function to reqRep (request reply).

py/farm_ng/core/event_service_tool.py
ethanrublee: I think if we want to subscribe to a protobuf stream, like OakSyncFrame, but only receive a sub field this could be accomplished with something like
```
path: "/oak0/sync_frame"
query: "type=OakSyncFrame&include=imu_packets&service_name=oak0"
```
Then the generic EventService can use the include_field=imu_packets to drop all fields that are not include in the query string, using the for field in message.ListFields(): if field not in include_list: message.ClearField(field) and this can be specific per subscription.

Alternatively, our services can do this with the URI path, and publish both an ``/oak0/sync_frame`` and an "/oak0/imu" message.

py/farm_ng/core/event_service.py
edgarriba: i like more explicit names -> `requestReply`

py/farm_ng/core/event_service_tool.py
ethanrublee: Haven't implemented this yet.  Suggest this is done after we have a few more tests and use cases.

py/farm_ng/core/event_service_recorder.py
```diff
@@ -0,0 +1,362 @@
...
+        recorder = EventServiceRecorder(_args.config_name, config_list=configs)
+        await recorder.subscribe_and_record(
+            _args.output, _args.extension, _args.max_file_mb
+        )
+
+    asyncio.get_event_loop().run_until_complete(job())
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
edgarriba, line 332: not in favor of having both the library definition and the code using it in the same place in order to keep clean of the tools needed by the library and what is needed to run it. For this reason they added the ability to define entrypoints in the `setup.py`. See: https://setuptools.pypa.io/en/latest/userguide/entry_point.html

edgarriba: i'd suggest to use some of the well supported community packages like `Typer` which has all the bells and whistles for building modern cli tools. See: https://typer.tiangolo.com/

py/farm_ng/core/event_service_recorder.py
```diff
@@ -75,15 +75,13 @@ def __init__(self, service_name: str, config_list: EventServiceConfigList) -> No
         self._clients: dict[str, EventClient] = {}
 
         # find the recorder config and create the clients
-        # TODO: why do not pass directly the config instead of the config list?
+        # Here we reuse a config for the subscriptions, from the recorder_config
+        # and create the clients for the other services (with port names)
```
edgarriba, line 79: passing the list can be confusing and error prone. I'd go for passing two arguments in the constructor `recorder_config` (a config ds with subscriptions only for later subscribe) and then the `config_list` containing the existing services with the  info needed to create the client (host, port, etc). I also feel that having a single `EventServiceConfig` that can be both for creating client and defining service is a bit counter intuitive to interpret.

py/farm_ng/core/event_service_recorder.py
```diff
@@ -197,6 +204,12 @@ def __init__(self, event_service: EventServiceGrpc) -> None:
         Args:
             event_service (EventServiceGrpc): the event service.
         """
+        parser = argparse.ArgumentParser()
```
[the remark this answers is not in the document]
ethanrublee: Ok this is reasonable for testing configuration parsing, however it does add quite some boilerplate. I think having a simple uniform pattern is important for the service classes; and i think service classes should just take the event_service as a single argument, and it should contain everything they need to get bootstrapped.

So a compromise here is to create a free function that goes from args -> MyStrongConfigType and call this in the constructor.  That function is then unit testable, and will have type hints for use inside the class.

py/farm_ng/core/event_service_recorder.py
ethanrublee: I'm not in favor of a new dependency for argument parsing, like typer.  The goal here is to have a simple uniform command line interface for all services, and argparse is built in to python.

python -m farm_ng.core.event_service_recorder --service-config config.json --service-name name_of_my_service

This allows for generic launching code with no specific knowledge of the implementation details of the service.

As far as entry points are concerned I'm not sure I understand the proposal, in this case it would be:

```
[project.scripts]
farm-ng-core-event-service-recorder = "farm_ng.core:event_service_recorder"
```

py/farm_ng/core/event_service_recorder.py
ethanrublee: This is designed to make the config.json structure easy to edit, with a single flat list of configuration sections.  Also, services will also have clients, so having a single EventServiceConfig that is either describing just a service, or a client, or a service and a client allows us to express this.

ethanrublee:
@strasdat @isherman this is ready for review.  This should have minimal impact on c++ code at the moment, but does move some protobufs from Sophus into farm_ng/core proto package (under a different namespace so can coexist).  

The logging macro changes were necessary when building in a simplified pybind environment (without cmake, and I'm not sure if these will impact the c++ experience or what a better fix is).

I'd like to merge this into main soon as its a large PR and relatively stable, would prefer smaller changes to fix it.

protos/farm_ng/core/image.proto
```diff
@@ -0,0 +1,42 @@
+
```
strasdat, line 1: Huh, I thought we wanted to move the whole sophus2 folder?

cpp/farm_ng/core/logging/logger.h
```diff
@@ -198,7 +199,7 @@ inline StreamLogger& defaultLogger() {
               rhs),                                                          \
           __FILE__,                                                          \
           __LINE__,                                                          \
-          __func__ FARM_MAYBE_VARGS(__VA_ARGS__)(__VA_ARGS__));              \
+          __func__);                                                         \
```
strasdat, line 202: This is removing context from all our logs and asserts in downstream dependencies if I'm. not mistaken.

We should either create a configurable log macro as we discussed or get it to work as is. (See passing CI in downstream project as documentation what build configuration is needed ...)

strasdat: Can you share the compile flags used for the python workflow when it is failing?

https://stackoverflow.com/questions/63740056/variadic-macros-with-zero-arguments-doesnt-compile-even-with-va-args

protos/farm_ng/core/image.proto
ethanrublee: not yet, this just the protobufs for now.  In a follow up we'll move the sophus2 folder.

strasdat (approved):
Approve - given coarse review.

Note that the assumption is that the protos are not suppose to work yet in downstream c++ code including sophus2. To work in downstream c++ projects, those protos need to be merged / harmonized with pre-existing protos (sophus2 and others).

py/farm_ng/core/event_service_tool.py
ethanrublee: Ok the conclusion is that each service should publish its own messages in a granular fashion such that subscriptions can be data effecient and simplify the event service/client pub/sub mechanism for now.  Later we can evaluate if we want to allow for subscription to sub-fields of protobufs.

py/farm_ng/core/event_service_recorder.py
ethanrublee: We'll revisit useability before we make an official release.

cpp/farm_ng/core/logging/logger.h
ethanrublee: compiling with ``--std=gnu++17`` worked rather than ``--std=c++17`` thank you @strasdat .  I've reverted these changes to logger.h.

py/farm_ng/core/event_service_recorder.py
ethanrublee: going to resolve this for now.

PR #97, opened 2023-07-28
For some reason log_level_ can't be printed easily.

ethanrublee:
For some reason in another code path we couldn't find the printer for log_level_.  Not sure.  @Hackerman342 can you past the traceback from the compile step?


Hackerman342:
If I isolate this change from the others in https://github.com/farm-ng/farm-ng-core/tree/many-hacks-to-logger, the build succeeds (see #98 )

ethanrublee:
#98 merged into event-service-client branch, see #93

PR #98, opened 2023-07-31
For some reason FARM_MAYBE_VARGS is causing issues

Hackerman342:
```
error: expected primary-expression before ')' token
        263 |           __func__ FARM_MAYBE_VARGS(__VA_ARGS__)(__VA_ARGS__));
```


ethanrublee:
I will merge this into event-service-client, and we can get help with getting to main brain later.

PR #99, opened 2023-08-04
Claire/tmp protos

lullabee:


protos/farm_ng/core/pose.proto
```diff
@@ -0,0 +1,32 @@
...
+
+// A position and orientation for an object or reference frame in 3D space
+message Pose3 {
+  Timestamp timestamp = 1;
+  // Translation from origin frame 
+  Vector3 translation = 2;
+  // Quaternion denoting orientation
+  Quaternion rotation = 3;
+
+  string frame_id = 4;
```
edgarriba, line 31: i assume this is the source frame ? missing also a key describing the destination frame of the transformation for this pose

edgarriba (commented):
i'd rather place this in `farm-ng-amiga` with all the other protos or at least under `farm_ng.geometry` namespace

PR #100, opened 2023-08-07
Claire/new protos

lullabee:


protos/farm_ng/core/pose.proto
```diff
@@ -0,0 +1,35 @@
...
+
+package farm_ng.core.proto;
+
+// A position and orientation for an object or reference frame in 3D space
+message Pose {
+  // Point_B = B_from_A * Point_A
+  Isometry3F64 B_from_A = 1;
+  string frame_A = 2;
+  string frame_B = 3;
+  // TODO: Velocity_B = B_from_A * Velocity_A
```
ethanrublee, line 28: @strasdat @stevenlovegrove help with this comment description in a follow up. We'd like to include velocity in some pose telemetry, and likely acceleration at some point, for more runtime flexibility in state estimation and control. A link to math documentation or simple inline explanation in proto would be good.

edgarriba: nitpick for naming based on protobuf style guide: https://protobuf.dev/programming-guides/style/#message-field-names
In `lower_sneak_case`  -> ```velocity_in_b = b_from_a * Velocity_a```

strasdat: I have just seen this comment:

For velocities, you typically you have *two* frames:

  1. The frame of the body in motion (e.g. "robot_base")
  2. And the frame of origin (e.g. "world").
  
(Sometimes there are even three: http://paulfurgale.info/news/2014/6/9/representing-robot-pose-the-good-the-bad-and-the-ugly - though the use-cases I dealt with two were sufficient. At least it is not clear to me when there is an advantages to distinguish between the frame of origin and the observer frame for classical mechanics.)

strasdat: (Well, I guess strictly speaking each time you have egocentric representations of velocities, such as velocity of the ROBOT in the ROBOT coordinate frame, you have a third implicit frame which is the world. Otherwise, such quantity will always be zero, since clearly there is no relative motion between frame X and itself by definition.
So, strictly speaking, we would need three frame to fully specify quantities such as velocities...)

PR #105, opened 2023-08-15
merge cmake+core+sophus2

strasdat:
1. Get sophus to build without proto support.
  - [x] locally
  - [x] in CI
2. Get sophus to build with proto/grpc support
  - [x] locally
  - [x] in CI
3. everything else


cmake/cmake/External_eigen.cmake
```diff
@@ -0,0 +1,10 @@
+ExternalProject_Add(eigen
+    GIT_REPOSITORY  https://gitlab.com/libeigen/eigen.git
```
ethanrublee, line 2: This could point to the local eigen/fmt/Expected submodules to not break anything.

strasdat: I'll try this in a follow-up!

cpp/sophus2/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+# Create examples make targets using ctest
```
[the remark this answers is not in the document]
strasdat: Yeah, we need to reason what the best structure is. I might make sense to keep sympy code which is mainly used for c++ code generation etc. close to the c++ code. Most folks you look for python code don't want that, so I might be actually good to have it more hidden.

PR #108, opened 2023-08-17
Modify stamp semantics in client / service

Hackerman342:


ethanrublee (commented):
I wonder if this should throw an exception rather than return None?

Suggest adding a unit test for coverage

py/farm_ng/core/stamp.py
```diff
@@ -54,3 +58,17 @@ def timestamp_from_monotonic(semantics: str, stamp: float) -> Timestamp:
         clock_name=get_host_name() + "/monotonic",
         semantics=semantics,
     )
+
+
+def get_stamp_by_semantics(event, semantics: str) -> float | None:
```
edgarriba, line 63: typing

edgarriba (approved):

edgarriba:
> I wonder if this should throw an exception rather than return None?
> 
> Suggest adding a unit test for coverage

+1 for tests; more a fan of letting the consumer of this function to decide whether to raise the exception

ethanrublee (approved):

PR #112, opened 2023-08-19
Bad subscription logic.

ethanrublee:


edgarriba (commented):
@ethanrublee If that’s breaking something we should add a test. I made that change based on https://beta.ruff.rs/docs/rules/superfluous-else-return/

ethanrublee:
This is a yield statement, so we need to continue the loop, so the
execution falls through and both are called.

PR #121, opened 2023-08-21
fix: proto style, attempt to correct velocity representation

strasdat:


ethanrublee:
Thank you for the post-merge review @strasdat - give me a few days to merge this in, as need to add some tests to pose_tree.py to update to this new convention, but I like the style and reference frame consistency.

ethanrublee (approved):
Added tests.

PR #124, opened 2023-08-22
add `EventServiceHealthMetrics`

edgarriba:


py/farm_ng/core/event_service_metrics.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def __call__(cls, *args, **kwargs):
+        if cls not in cls._instances:
+            cls._instances[cls] = super().__call__(*args, **kwargs)
+        return cls._instances[cls]
+
+
+class EventServiceHealthMetrics(metaclass=Singleton):
+    def __init__(self) -> None:
+        self.service_counts: dict[str, int] = {}
```
ethanrublee, line 17: Question should this just hold an instance of the protobuf we intend to publish?

py/farm_ng/core/event_service.py
```diff
@@ -178,7 +187,25 @@ async def serve(self) -> None:
...
+    async def publish_health_metrics(self) -> None:
+        """Publishes the health metrics of the service."""
+        while True:
+            # decide later a better message type
+            health_metrics = {}
+            for uri_path, counts in self._metrics.service_counts.items():
+                health_metrics[uri_path] = counts
+
+            await self.publish(
+                path=f"{self.config.name}/health",
```
ethanrublee, line 205: We don't pre-pend the config.name when sending.  should just be path="/health"

py/farm_ng/core/event_service.py
```diff
@@ -420,6 +447,7 @@ async def _send_raw(
         # so that the client can know that the message was not sent by checking
         # the count of the URI.
         self._counts[uri.path] = count + 1
+        self._metrics.service_counts[uri.path] = count + 1
```
ethanrublee, line 450: Here i imagine the metric "key" to be f"{uri.path}/send_count", or similar.  Likely metrics are simple key value pairs, with a set of relatively constrained value types.

Good inspiration can be found from Prometheus: https://prometheus.io/docs/concepts/metric_types/

Imagine as a developer I want to create an eventservice metrics to prometheus bridge, to dashboard the health of the system.

py/farm_ng/core/event_service_metrics.py
edgarriba: done using `struct_pb2.Struct`

py/farm_ng/core/event_service.py
```diff
@@ -431,13 +433,29 @@ async def _send_raw(
...
+            self._metrics.data[uri_path_send_clients] = 0
+
+        self._metrics.data[uri_path_send_clients] += num_clients
+
+        # update the metrics with the timestamp of the last message
+        # to compute the rate of the messages.
+        if uri.path not in self._metrics.data_tmp:
+            self._metrics.data_tmp[uri.path] = []
+
+        self._metrics.data_tmp[uri.path].append(timestamps[-1].stamp)
```
Hackerman342, line 454: Suggest to use [`get_stamp_by_semantics_and_clock_type`](https://github.com/farm-ng/farm-ng-core/blob/main/py/farm_ng/core/stamp.py#L68) rather than indexing by order in a lsit

py/farm_ng/core/event_client.py
```diff
@@ -251,8 +255,12 @@ async def request_reply(
...
 
         # get the current count and increment it
-        count: int = self._counts.get(path, 0)
-        self._counts[path] = count + 1
+        path_received = f"{path}/receive_count"
+        if path_received not in self._metrics.data:
+            self._metrics.data[path_received] = 0
+
+        count: int = int(self._metrics.data[path_received])
+        self._metrics.data[path_received] = count + 1
```
Hackerman342, line 263: Suggest to look at the latency between `SERVICE_SEND` & `CLIENT_RECEIVE` stamps (in host monotonic clock) here and especially in the subscribe method. This has proven to be a very important metric and will be a quick indicator of issues in our pipeline.

This will not be meaningful if service and client are on a different host (e.g. using a dev station as client) so a first version could only report IFF the service and client are on the same host. And we could add a host/client monotonic clock synchronization to the API in the future.

py/farm_ng/core/event_service.py
edgarriba: done

py/farm_ng/core/event_service_metrics.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # private methods
+
+    def _compute_avg_latency(self, stamps: deque[float]) -> float:
+        if len(stamps) < self.MIN_NUM_ELEMENTS:
+            return 0.0
+
+        latencies: list[float] = [
+            stamps[i] - stamps[i - 1] for i in range(1, len(stamps))
+        ]
```
Hackerman342, line 73: This isn't really latency. This is just the dt between packets / the inverse of the rate. Latency would be the time between service send and client receive.

Hackerman342: Suggest to rename until it is actually latency

edgarriba: Got it

py/farm_ng/core/event_client.py
```diff
@@ -91,7 +93,7 @@ def __init__(self, config: EventServiceConfig) -> None:
         self._logger: logging.Logger = logging.getLogger(f"{self.config.name}/client")
 
         # the number of messages sent
-        self._counts: dict[str, int] = {}
+        self._metrics = EventServiceHealthMetrics()
```
ethanrublee, line 96: Shouldn't this use the singleton for metrics?  This will allow any EventService that has EventClients to publish metrics only measurable by clients.

ethanrublee: Or is this a singleton due to the base class?

edgarriba: it's a singleton

py/farm_ng/core/event_client.py
ethanrublee: This is really request /reply metrics in this location.  Above in subscirbe we can compute the latency between SERVICE_SEND and CLIENT_RECEIVE (as they are in the timestamps field) and this is the metric we are quite interested in, in addition to count and frequency.

ethanrublee (approved):

PR #125, opened 2023-08-24
Rename vel vecs in Pose proto

Hackerman342:


protos/farm_ng/core/pose.proto
```diff
@@ -26,12 +26,11 @@ message Pose {
...
+  Vec3F64 linvel_of_b_in_a = 4;
 
-  // linear velocity of body a (with respect to body b) in frame b.
-  Vec3F64 linvel_of_a_in_b = 4;
-
-  // angular velocity of body a (with respect to body b) in frame a.
-  Vec3F64 angvel_of_a_in_a = 5;
+  // angular velocity of body b (with respect to body b) that exists in frame a.
+  Vec3F64 angvel_of_b_in_a = 5;
 }
```
Hackerman342, line 35: We will remove velocity from pose and create another message:
```
message Velocity {
str body_frame = 1; // the name of the frame of the body
vec3 linvel =2; // egocentric linear velocity x,y,z
vec3 angvel = 3; // egocentric angular velocities x,y,z
}
```

Hackerman342: No actually we still need parent frame here...

protos/farm_ng/core/pose.proto
```diff
@@ -26,12 +26,11 @@ message Pose {
   string frame_a = 2;
   string frame_b = 3;
 
+  // linear velocity of body b (with respect to body b) that exists in frame a.
```
Hackerman342, line 29: From convo w/ @strasdat 
Change comment to something more like:
// Linear velocity of frame b with respect to frame a
// Egocentric velocity of body b with regard to frame a.
// e.g., if b is 'robot' and 'a' is world,this is the linear velocity of the robot (x is moving forward, y left, z up)
// If 'b' is a tool and 'a' is the robot, the robot velocity is not included. Only the actuation of the tool relative to robot's body, but in the tool's axes.

Maybe change name to `linvelv_of_b_in_b`. There are really 3 reference frames, so the name is hard to nail down here

Hackerman342:
@strasdat please take a look at the updated comments after our discussion. I am going to merge once this passes CI to unblock downstream usage of these changed proto names. We can further clarify / iterate on the comments as needed.

PR #128, opened 2023-08-26
Gui/covariance proto

guilhermedemouraa:
Protobuff definition for state covariances


ethanrublee:
Recommend closing.  I think you can use the Vec3F64 or VecXF64 protobuf types for this use case:
https://github.com/farm-ng/farm-ng-core/blob/main/protos/farm_ng/core/linalg.proto#L78

PR #131, opened 2023-08-28
Match setup.py __version__ to setup.cfg version

Hackerman342:


edgarriba (commented):
I would follow this pattern: https://github.com/farm-ng/farm-ng-amiga/blob/main/py/farm_ng/oak/__init__.py

to avoid maintaining from two different locations

Hackerman342:
> I would follow this pattern: https://github.com/farm-ng/farm-ng-amiga/blob/main/py/farm_ng/oak/__init__.py
> 
> to avoid maintaining from two different locations

Any reason not to require >=3.8 since ,=3.7 are all EOL ?

Hackerman342:
Seems this was unrelated to the problem at hand and that `__version__` in setup.py is not visible from an API perspective

PR #132, opened 2023-08-29
Event service backend

edgarriba:


py/farm_ng/core/event_service_backend.py
```diff
@@ -0,0 +1,186 @@
...
+
+    Attributes:
+        service_config: The configuration for the service.
+        clients: A map of service name to EventClient.
+        subscriptions: A map of service_path to SubscribeRequest.
+    """
+
+    def __init__(
+        self,
+        service_config: EventServiceConfig,
```
ethanrublee, line 30: Suggest not taking a service_config, just a pre-filtered config list

CarlSaganPhD (approved):

PR #133, opened 2023-08-29
Support --config-name for recorder start command

Hackerman342:


ethanrublee:
Please add a test for this.

ethanrublee (changes requested):
Please add a test case for this.

py/farm_ng/core/event_service_recorder.py
```diff
@@ -357,9 +368,13 @@ async def job():
 def client_start_command(_args):
     config_list, service_config = load_service_config(args)
     loop = asyncio.get_event_loop()
+    config_name: str = args.config_name
 
     async def job():
-        reply = await EventClient(service_config).request_reply("start", config_list)
+        reply = await EventClient(service_config).request_reply(
+            f"start/{config_name}",
```
ethanrublee, line 375: Yes makes sense!

Hackerman342: :sweat_smile:

guilhermedemouraa (approved):
Beautiful!

ethanrublee:
For test, you add a simple change here to cover the new request/response api: py/tests/_asyncio/test_event_service_recorder.py

Hackerman342:
@ethanrublee I've been working on this. Maybe I'm missing something, but I don't see it as a simple tweak to the existing tests. There is no test infrastructure for the `RecorderService`, only the `EventServiceRecorder`.

ethanrublee:
Ok bummer.  I see that now.

ethanrublee:
We could follow up with a test I guess then.

ethanrublee (approved):

PR #136, opened 2023-08-30
add test for recorder service

edgarriba:


py/tests/_asyncio/test_event_service_recorder.py
```diff
@@ -78,3 +79,47 @@ async def test_record(
...
+        tmp_path: Path,
+        event_service_config_list: EventServiceConfigList,
+        event_service: EventServiceGrpc,
+    ) -> None:
+        # reset the counts
+        event_service.reset()
+        event_service.config.args.extend(["--data-dir", str(tmp_path)])
+
+        # create the recorder service
+        recorder_service = RecorderService(event_service)
```
edgarriba, line 97: @ethanrublee @Hackerman342 i propose to pass data-dir as parameter to the RecorderService so that its more natural to use at least for testing instead of having an argparser in the constructor

Hackerman342: Yeah I'm good with that. I was surprised to see the arg parse in the constructor. I had thought we needed an event service created with the record_default config, but I wasn't really sure

ethanrublee: For now I prefer not to change this api, as we start it from a config file in production, and don't see much of a need to start it some other way. This way we test how we run.

ethanrublee (approved):

ethanrublee (approved):

Hackerman342:
Thank you @edgarriba :tada:

PR #143, opened 2023-09-05
EventService request reply interface simplification.

ethanrublee:


py/farm_ng/core/uri.py
```diff
@@ -70,7 +70,7 @@ def get_authority() -> str:
     return get_host_name()
 
 
-def make_proto_uri(path: str, message: Message) -> uri_pb2.Uri:
+def make_proto_uri(path: str, message: Message, service_name=None) -> uri_pb2.Uri:
```
edgarriba, line 73:
```suggestion
def make_proto_uri(path: str, message: Message, service_name: str | None = None) -> uri_pb2.Uri:
```

edgarriba (approved):

PR #147, opened 2023-09-06
Support custom clock in logger

isherman:


cpp/farm_ng/core/logging/logger.h
```diff
@@ -40,19 +40,23 @@ std::string stringFromLogLevel(LogLevel level);
...
 
   // Set the runtime log level
   void setLogLevel(LogLevel level);
-
   LogLevel getLogLevel();
 
+  // A type-erased generator of timestamp strings
+  struct LogClock {
+    std::function<std::string()> now;
+  };
```
wylderkeane, line 55: Any reason to not just do `using LogClock = std::function<std::string()>;` ?

cpp/farm_ng/core/logging/logger.cpp
```diff
@@ -58,4 +54,17 @@ void StreamLogger::writeHeader(
 void StreamLogger::write(std::string const& str) { std::cerr << str; }
 void StreamLogger::flush() { std::cerr << std::endl; }
 
+const std::string StreamLogger::kDefaultHeaderFormat =
+    "[FARM {text} in {file}:{line}]\n";
+
+const StreamLogger::LogClock StreamLogger::kDefaultLogClock =
+    StreamLogger::LogClock{.now = []() {
+      auto now = std::chrono::system_clock::now();
```
wylderkeane, line 62: Since we're defaulting to system_time, shouldn't we also update the vns_app to initialize this with the `MissionClock::now()` call?

isherman: This is in reference to a downstream project; please see its respective repository for that PR

cpp/farm_ng/core/logging/logger.h
isherman: I thought about it, and liked this API better in terms of the type names and how it reads at the call-site. You might be right; I could go either way.

strasdat (approved):

PR #149, opened 2023-09-06
proto definition for UKF convergence

guilhermedemouraa:
I couldn't find an existing protobuff definition including a `bool` type message, so I created this one for debugging my code. I am happy to place it somewhere else or use any equivalent


edgarriba:
mind closing this

PR #169, opened 2023-09-27
Fix PyPi release action

Hackerman342:


edgarriba (commented):
don't we need a arm64 runner too?

Hackerman342:
Split into #170 & #171

PR #171, opened 2023-09-28
Try revive macos-latest build -- fails for pybind code

Hackerman342:
Attempt at reviving the ability to pip install on a `macos-latest` runner. Spun off from #169, which was split into two so we can separate the #170  from the more difficult problem of reviving the macos build.

### DIFF
--- a/.github/workflows/ci_python.yml
+++ b/.github/workflows/ci_python.yml
@@ -28,8 +28,10 @@ jobs:
       run: |
         # Install GCC on macOS
         brew install gcc@11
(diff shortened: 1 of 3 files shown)


Hackerman342:
Check the action output for the error for the latest attempt, but it seems to me the issue is the `macos-latest` action is using clang, despite attempts to force the use of `gcc`. This causes a failure due to the `-fconcepts` build arg

ethanrublee:
@Hackerman342 this now builds and tests pass on mac for the python code.

ethanrublee (approved):

Hackerman342:
Awesome!

PR #179, opened 2023-10-08
feat: plot nav improvements

strasdat:
[Screencast from 2023-10-09 23-35-12.webm](https://github.com/farm-ng/farm-ng-core/assets/584721/4eb6ef2e-776a-4c75-9a2f-95edb211d2b2)

Stand-alone example (without need of grpc or c++):

```
cargo run -bin plotter_example
```

### DIFF
--- a/rs/plotting/src/graphs/common.rs
+++ b/rs/plotting/src/graphs/common.rs
@@ -34,6 +34,33 @@ impl Color {
         }
     }
 
+    pub fn dark_red() -> Self {
+        Color {
+            r: 0.5,
+            g: 0.0,
+            b: 0.0,
+            a: 1.0,
+        }
+    }
+
+    pub fn  dark_green() -> Self {
+        Color {
+            r: 0.0,
+            g: 0.5,
+            b: 0.0,
+            a: 1.0,
+        }
+    }
+
+    pub fn  dark_blue() -> Self {
+        Color {
+            r: 0.0,
+            g: 0.0,
+            b: 0.5,
+            a: 1.0,
+        }
+    }
+
     pub fn new() -> Self {
         Color {
             r: 0.5,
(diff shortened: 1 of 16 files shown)


rs/plotting/src/graphs/common.rs
```diff
@@ -44,6 +44,34 @@ impl Color {
...
+pub struct RegionF64 {
+    pub min: f64,
+    pub max: f64,
+}
+#[derive(Clone, Debug, Copy)]
+pub struct OrdinateBounds {
+    pub largest: f64,
+    pub len: f64,
+    pub data_driven: bool,
+}
```
isherman, line 57: Documentation of these members could be helpful.

rs/plotting/src/graphs/common.rs
```diff
@@ -44,6 +44,34 @@ impl Color {
...
+
+impl OrdinateBounds {
+    pub fn from_len_and_max(len: f64, max: Option<f64>) -> Self {
+        OrdinateBounds {
+            largest: max.unwrap_or_default(),
```
isherman, line 62: Should we call the member `max` instead of `largest`?

rs/plotting/src/graphs/common.rs
```diff
@@ -44,6 +44,34 @@ impl Color {
...
+#[derive(Clone, Debug, Copy)]
+pub struct RegionF64 {
+    pub min: f64,
+    pub max: f64,
+}
+#[derive(Clone, Debug, Copy)]
+pub struct OrdinateBounds {
+    pub largest: f64,
+    pub len: f64,
+    pub data_driven: bool,
```
isherman, line 56: Instead of a separate data driven member should `largest` just be an Optional?

rs/plotting/src/plotter_gui/mod.rs
```diff
@@ -165,127 +170,211 @@ impl eframe::App for PlotterGuiState {
...
-                                        points[1].push(egui::plot::PlotPoint::new(*x, y.1));
-                                        points[2].push(egui::plot::PlotPoint::new(*x, y.2));
+                                        if x > &data_driven_largest_x {
+                                            data_driven_largest_x = *x;
+                                        }
+
+                                        let max_y = y.0.max(y.1.max(y.2));
+                                        if max_y > data_driven_largest_y {
+                                            data_driven_largest_y = max_y;
+                                        }
```
isherman, line 315: I have a vague sense that we're interleaving UI code and UI model code more than would be ideal, at least in this `.show()` lambda, but I'm not very familiar with this area of the code.

isherman (approved):

rs/plotting/src/plotter_gui/mod.rs
strasdat: Good question. Ideally we have more general predicates and some sort of DSL, so we can formulate:

```
x-max-bound = MAX_X(DATA)
```

But I do not think it is worth the time working on something like this now, even though it would be cleaner.

rs/plotting/src/graphs/common.rs
```diff
@@ -44,6 +44,34 @@ impl Color {
     }
 }
 
+#[derive(Clone, Debug, Copy)]
+pub struct RegionF64 {
+    pub min: f64,
+    pub max: f64,
+}
+#[derive(Clone, Debug, Copy)]
+pub struct OrdinateBounds {
```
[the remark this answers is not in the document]
strasdat: done

rs/plotting/src/graphs/common.rs
strasdat: done

rs/plotting/src/graphs/common.rs
strasdat: done

strasdat:
TODO: YCoordinatesBounds in data driven mode shall not be symmetric but set an offset

strasdat:
A Vec3CurveWithConfInterval example needs to be included for c++ and rust.

PR #180, opened 2023-10-08
Fix protobuf warnings

ethanrublee:


ethanrublee:
For some reason I get a commit from Ian here, not sure.  I'll try rebasing against main, but did that commit not make it into main?

strasdat (approved):

PR #192, opened 2023-11-10
Build python wheels for aarch64

edgarriba:


ianBtesser (commented):
Cool

.github/workflows/pypi-release.yml
```diff
@@ -17,23 +17,24 @@ jobs:
...
-      #  with:
-      #    platforms: all
+      - name: Set up QEMU
+        if: runner.os == 'Linux'
+        uses: docker/setup-qemu-action@v3
+        with:
+          platforms: all
       - name: Install dependencies
-        run: python3 -m pip install --upgrade pybind11 setuptools wheel cibuildwheel==2.16.2
+        run: python3 -m pip install --upgrade pybind11 setuptools wheel cibuildwheel==2.3.1
```
mihelich, line 26: Is the constraint `cibuildwheel==2.3.1` just for build consistency, or is there a specific reason? Worth adding a comment.

mihelich (approved):
Thanks for solving this!

.github/workflows/pypi-release.yml
edgarriba: that's a mistake actually :)

PR #194, opened 2023-11-28
fix: check near

strasdat:
- implement FARM_NEAR according to best practices 
  * in a symmetric form such that ``relativeCloseness(x, y) == relativeCloseness(y, x)``
  * detecting NANs and PANIC in that case
- adding in FARM_ABS_NEAR if absolute comparison is needed instead: |x-y| <= thr
- fix newly detected bug (test was erroneously passing for a NAN value)

Fixes / clarifies issue discussed in this comment: https://github.com/farm-ng/farm-ng-core/pull/89#issuecomment-1604558108


cpp/farm_ng/core/logging/check_near.h
```diff
@@ -0,0 +1,210 @@
...
+  return CheckNear<TScalar, TScalar2>::check(
+      true, lhs, lhs_cstr, rhs, rhs_cstr, thr);
+}
+
+}  // namespace farm_ng
+
+/// If it is false that `lhs` is near `rhs` according to threshold `thr`, print
+/// formatted error message and then panic.
+///
+/// `lhs` and `rhs` are near, if relativeCloseness(lhs, rhs) < thr.
```
strasdat, line 173: Document that threshold shall be in [0; 2].

cpp/farm_ng/core/logging/check_near.h
```diff
@@ -0,0 +1,222 @@
...
+
+template <class TPoint, class TPoint2>
+struct CheckNear<
+    TPoint,
+    TPoint2,
+    std::enable_if_t<
+        std::is_arithmetic_v<TPoint> && std::is_arithmetic_v<TPoint2>,
+        void>> {
+  static Expected<Success> check(
+      bool check_relative,
```
isherman, line 59: I don't really like this as a bool argument, makes the call-site hard to read

isherman: Suppose I can deal with it if it's not part of the public API

cpp/farm_ng/core/logging/check_near.h
```diff
@@ -0,0 +1,222 @@
...
+
+#include <Eigen/Core>
+
+namespace farm_ng {
+
+/// Checks whether two floating point numbers x, y are close to each other.
+///
+/// It outputs:  min(|x-y|, |x-y| / max(|x|, |y|))
+///
+/// The output it between 0.0 and 2.0:
```
isherman, line 27: nit "is"

cpp/farm_ng/core/logging/check_near.h
```diff
@@ -0,0 +1,222 @@
...
+template <class TScalar, class TScalar2>
+inline Expected<Success> checkRelativeNear(
+    TScalar lhs,
+    char const* lhs_cstr,
+    TScalar2 rhs,
+    char const* rhs_cstr,
+    double thr) {
+  if (!((thr >= 0.0) && (thr <= 1.0))) {
+    FARM_WARN(
+        "FARM_ASSERT_NEAR: The threshold shall be in [0.0, 1.0], but we got "
```
isherman, line 167: Do you want the "FARM_ASSERT_NEAR" in this error string?

isherman (commented):
I'm not crazy about the names in the public API: "FARM_ASSERT_NEAR" and "FARM_ASSERT_ABS_NEAR". 

I think it's because I would use the "FARM_ASSERT_ABS_NEAR" whether or not I was dealing with numbers that were "near" each other. That's the use case that [originally confused me](https://github.com/farm-ng/farm-ng-core/pull/89/files#diff-0f01971348bd1c45be16370a409ab44667bd7dc55bbfdda10dd4433b390925eaR81-R83) about the API.

Here's another example:
```cpp
double t_a = getTimestamp("a");
double t_b = getTimestamp("b");

// Sanity check that event a occurred within an hour of event b
FARM_ASSERT_ABS_NEAR(t0, t1, 60 * 60);
```

Could we follow the [convention used in Catch2](https://github.com/catchorg/Catch2/blob/devel/docs/comparing-floating-point-numbers.md) and name them `FARM_ASSERT_WITHIN_ABS` and `FARM_ASSERT_WITHIN_REL`?

strasdat:
@isherman thanks for the naming suggestion, I adopted them as proposed.

isherman (approved):

PR #196, opened 2023-12-08
Add header messages (including latched) to EventsFileWriter

Hackerman342:


edgarriba (commented):
can you add tests ?

py/farm_ng/core/event_service_recorder.py
```diff
@@ -144,6 +165,14 @@ async def subscribe(
         event: event_pb2.Event
         payload: bytes
         async for event, payload in client.subscribe(subscription, decode=False):
+            if event.uri in header_uris:
+                # Handle header messages - typically these will be metadata or calibrations
+                # This assumes they are only published once (likely with latching=True),
+                # as we break the subscription after the first header message is received
+                self.logger.info("header received: %s", event)
+                self.header_deque.append((event, payload))
```
ethanrublee, line 173: This seems risky. Can you replace a header message if the uri is the same?

py/farm_ng/core/events_file_writer.py
```diff
@@ -135,10 +139,50 @@ def _increment_file_idx(self) -> None:
...
+            payload: Payload to write.
+            write: If True, write the header message to the file. Defaults to False.
+        """
+        if not isinstance(event, Event):
+            error_msg = f"event must be Event, not {type(event)}"
+            raise TypeError(error_msg)
+        if not isinstance(payload, bytes):
+            error_msg = f"payload must be bytes, not {type(payload)}"
+            raise TypeError(error_msg)
+        self._header_msgs.append((event, payload))
```
ethanrublee, line 165: Here I would suggest making header msgs be a dict based on the uri, so that for any header message uri we only support one message so that it can't grow in an unbounded way

ethanrublee (changes requested):

py/farm_ng/core/events_file_writer.py
Hackerman342: Yeah great idea. This simplifies the risky handling you noted above

Hackerman342: OK @ethanrublee current intended behavior:

- `RecorderService` allows for overwriting header fields with a matching `Uri` (dict key is `uri_to_string(Uri)`)
- `EventsFileWriter` does NOT allow for overwriting fields with a matching `Uri` (same key)

This way once you send the `start` command to the `RecorderService`, the headers the `RecorderService` has on hand are added to the log file and all subsequent file splits. If any headers with a new `Uri` are received while logging, they are written to the current file split immediately and are added to the header in all future file splits. Any headers on an existing Uri are ignored by the `eventsFileWriter`. This prevents header mismatch between log file splits (with the exception of added header messages).


After that logging session is complete (send `stop` command or another `start` and `EventsFileWriter` is closed), the next `EventsFileWriter` logging session will use any updated header messages sent to the `RecorderService` during the previous logging session or since its conclusion.


Last thing is for the latched topics that we use as headers (e.g. `/oak0/calibration`). The topic + service names are defined in the config for the `RecorderService`, but the `RecorderService` never actually sees that message or stores it with the manually defined headers (like the key-value pair `metadata` in the camera app). Rather the `EventServiceRecorder` receives it from the latching publisher when it is created by the `RecorderService` `start` command. It passes the message to the `EventsFileWriter` to add to the header messages for this logging session. We could pass these up to the `RecorderService`to maintain as well if we want, but it is not currently doing that.

Please let me know:
1. If this is reasonable behavior
2. If we should pass the latched messages up to the `RecorderService` to maintain

py/farm_ng/core/event_service_recorder.py
Hackerman342: Now you cannot. See other conversation thread.

py/farm_ng/core/event_service_recorder.py
```diff
@@ -95,6 +101,9 @@ def __init__(self, service_name: str, config_list: EventServiceConfigList) -> No
         self.record_queue: asyncio.Queue[tuple[event_pb2.Event, bytes]] = asyncio.Queue(
             maxsize=self.QUEUE_MAX_SIZE,
         )
+        self.header_deque: deque[tuple[event_pb2.Event, bytes]] = deque()
+        if header_msgs:
```
edgarriba, line 105:
```suggestion
        if header_msgs is not None:
```
to avoid someone passing a bool or string and append that. Maybe we should check on the type of the header message bedfore appending to the queue

py/farm_ng/core/events_file_writer.py
```diff
@@ -135,10 +141,54 @@ def _increment_file_idx(self) -> None:
         """Increment the file index."""
         self._file_idx += 1
 
+    @property
+    def header_msgs(self) -> dict[str, tuple[Event, bytes]]:
+        """Return the list of header messages.
```
edgarriba, line 146:
```suggestion
        """Return the list of header messages.
 
```

py/farm_ng/core/events_file_writer.py
```diff
@@ -135,10 +141,54 @@ def _increment_file_idx(self) -> None:
...
+        Returns:
+            dict[str, tuple[Event, bytes]]: Dictionary of header messages.
+                key: string representation of the uri
+                value: tuple of event and payload
+        """
+        return self._header_msgs
+
+    def add_header_msg(self, event: Event, payload: bytes, write: bool = False) -> None:
+        """Add a header message, and optionally writes it to the file.
```
edgarriba, line 155:
```suggestion
        """Add a header message, and optionally writes it to the file.

```

edgarriba (approved):
LGTM. Question/suggestion: how can we read/parse easily header messages ?

py/farm_ng/core/events_file_writer.py
Hackerman342: Adding this to pre-commit hooks in https://github.com/farm-ng/farm-ng-core/pull/201

py/farm_ng/core/event_service_recorder.py
Hackerman342: Yeah good call. We check in `EventsFIleWriter`, but better to check further upstream

ethanrublee:
I still feel like this could be simpler.

I can imagine latched topics can generate new data mid recording pretty easily, which would cause this to fail, and a programmer or user to scratch their head.

Also if new meta data is added after a split, header messages will be inconsistent across log files. When is this likely to be needed?

As a user how do I edit the meta data after it's recorded?

I think for latched topics, if we can extend the query string on the sender side, then the latched topics can easily and transparently be carried across log splits with no additional api.

Could meta data just be of the convention /metadata/foo and be a latched topic?

And then we keep the semantics simple and really don't need to change the APIs?

Hackerman342:
@ethanrublee 
> I still feel like this could be simpler.

Yeah I'd like it to be too.

> I can imagine latched topics can generate new data mid recording pretty easily, which would cause this to fail, and a programmer or user to scratch their head.

I don't think this does fail .Or maybe I'm misunderstanding what you mean by fail? There's no special casing for latched topics anymore. The only special case is a white list of URIs to consider as header messages, whether it is latched or not. That message comes in, we try to add it as a header, and we write it to the log either way. If we already created a header with that URI in this recording, then the old header is not overwritten. So either way, the new message received is still included in the log.

```
def subscribe():
        async for event, payload in client.subscribe(subscription, decode=False):
            if uri_to_string(event.uri) in header_uris:
                # Handle header events - typically these will be metadata or calibrations
                self.header_deque.append((event, payload))
            try:
                self.record_queue.put_nowait((event, payload))
            except asyncio.QueueFull:
                self.logger.warning("Queue full, dropping event %s", event)
                continue
```

> Also if new meta data is added after a split, header messages will be inconsistent across log files. When is this likely to be needed?

Likely never. Just thinking through edge cases.

> As a user how do I edit the meta data after it's recorded?

We could make a python script with an example for this in `farm-ng-amiga` when this comes up as a need

> I think for latched topics, if we can extend the query string on the sender side, then the latched topics can easily and transparently be carried across log splits with no additional api.

Oh I like this. Instead of a whitelist of subscribed topic URIs, we check for `/header` as the final component of the Uri path. That simplifies that logic and can remove the change to the proto definition in this PR

> Could meta data just be of the convention /header/foo and be a latched topic?
>
> And then we keep the semantics simple and really don't need to change the APIs?

This would require creating a separate service / publisher for the recording app just to publish metadata. I think it makes more sense to keep it as a client and use the request/reply API, but if you'd prefer that direction I can do that,

Hackerman342:
@edgarriba 
> LGTM. Question/suggestion: how can we read/parse easily header messages ?

Good idea. I added a method for this in: https://github.com/farm-ng/farm-ng-core/pull/196/commits/e790025f6295bb7ce46fa9e9a3837869343ff21c

ethanrublee (approved):

PR #197, opened 2023-12-08
feat: image transform with uv index

strasdat:


isherman (approved):
Looks good. Could be collapsed w/ `makeFromTransform()` since it's a strict superset of the functionality but just a thought.

PR #198, opened 2023-12-12
Cubic BSpline dx

strasdat:
- added Cartesian version of the Cubic BSspline (as oppose to the more general LieGroup version), since it is easier to reason about it etc. (Also first studying the Cartesian version might help understanding the Lie Group one...)
 - added analytic Jacobians wrt to the control points for the Cartesian Cubic BSspline 
 -  added test coverage
    * for the correctness of the Jacobian
    * that the Cartesian version of the spline and the Lie Group on behave identical (if LieGroup = Translation<...>)
 - refactoring
    * move interp in its own dedicated directory
    * split the massive spline.h file into multiple more consumable chunks
    * more precise naming, e.g. CubicBSpline instead of BasisSpline ...


cpp/sophus/interp/spline/details/group_bspline_segment.h
```diff
@@ -0,0 +1,286 @@
+// Copyright (c) 2011, Hauke Strasdat
```
strasdat, line 1: Moved and light refactored (more precise names) from https://github.com/farm-ng/farm-ng-core/blob/038a0c3f6ed6be19b95fa6cc2e3e0153b474c31d/cpp/sophus/lie/interp/spline.h

cpp/sophus/interp/spline/details/cubic_basis.h
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2011, Hauke Strasdat
```
strasdat, line 1: Moved and light refactored (more precise names) from https://github.com/farm-ng/farm-ng-core/blob/038a0c3f6ed6be19b95fa6cc2e3e0153b474c31d/cpp/sophus/lie/interp/spline.h

PR #202, opened 2023-12-19
Experiments with cmake and packaging

isherman:
# Overview

- Uses CMakePresets.json to define common build configurations as presets

  - Includes presets for clang/gcc, debug/release, clang-tidy asan, tsan, msan
  - Persistent build directories per preset
  - Great Vscode integration
  - Developer can add CMakeUserPresets.json for additional configurations

- Separate package definition from development environment

  - Modern cmake best practices encourage us to separate build/packaging requirements (used by downstream consumers)
    from the development environment (used by developers)
  - For example:

    - Our CMakeLists.txt should not specify where to find dependencies, only that they are found by `find_package`
    - Our CMakeLists.txt should not specify build flags, only target requirements (e.g. `target_compile_features`)

- Leverage vcpkg as a package manager

  - This setup does not require use of a package manager, but recommends it and uses it in CI
  - Vcpkg has been easy to setup and the semantics feel minimal and complete
  - To me, use of a package manager feels like the only way to address potential ODR violations.
    You need a centralized place to reason about package versioning compatibility across the entire dependency tree.
  - Eliminates the need for our custom C++ "virtualenv" setup
  - Open-source contributors contribute patches that improve packaging of popular libraries (e.g. support for static linking)

- Easily switch between development / released / system versions of packages

  - Switching to a development version relies on a few lines of code in `cmake/farm_ng_dev_packages.cmake` and one-line edits to `CMakePresets.json` and `vcpkg.json`
    This seems to be working so far -- it is demonstrated here with sophus, Eigen, and CLI11. I also played with CMake's "Dependency Providers" feature which allows more customization if necessary.
  - Switching to a system version requires a one-line edit to `vcpkg.json`, and then find_package() should resolve to a system installed version.

- Move towards Sophus as separate project

  - For simplicity it is still vendored in here, but can easily be made a submodule or pulled in with vcpkg

- Optional devcontainer.json

  - Dockerized development environment is optional but not required
  - Better reproducibility for CI
  - One-click Github Codespaces integration

- Better cross-platform support

  - OSX should work, clear path to Windows
  - Clear path to support cross-compilation

- Begins to explore refactoring/cleanup of farm_ng module macros in `cmake/`
  - There are some new Cmake APIs and recommended patterns we can take advantage of

# Instructions

```bash
# Open a terminal **without** the farm-ng venv activated
# Should print 'OK':
printenv CMAKE_PREFIX_PATH || echo "OK"

# Ensure you have a reasonably up-to-date cmake (tested with 3.25) and ninja (tested with 1.10)
cmake --version
ninja --version

# Update submodules
git submodule update --init --recursive

# Bootstrap vcpkg
./vcpkg/bootstrap-vcpkg.sh

# Configure
# Note: Builds dependencies from source on first invocation, may take 5-10mins
cmake --preset clang-debug

# Build
cmake --build --preset clang-debug

# Test (one test fails on my machine)
ctest --preset clang-debug

# Install
mkdir /tmp/farm-ng-prefix
cmake --build --preset clang-debug --target install -DCMAKE_PREFIX_PATH=/tmp/farm-ng-prefix

# Try the same workflows in Vscode
# Select a configure preset such as "Clang RelWithDebInfo", then build and run as usual
code farm_ng_core.code-workspace

# Optionally, you can try the devcontainer workflow in Vscode by clicking "Reopen in Container" when prompted
# - Note, this will not re-use the same vcpkg cache, so the first build will take more time
# - Note, if you select a configure preset in the devcontainer that you've already built on your host machine you will need to "CMake: Delete Cache and Reconfigure"
```

# FAQ

**Q: How does this work for repositories downstream of farm-ng-core? What about consuming closed-source packages?**

A: Largely the same. It will help if we define a vcpkg "port" for each farm-ng C++ package ([an example](https://github.com/northwindtraders/vcpkg-registry/blob/main/ports/beicode/portfile.cmake)). We can then host these ports in a farm-ng vcpkg registry, which is just a (optionally private) Github repository ([an example](https://github.com/northwindtraders/vcpkg-registry/tree/main)). I have not tried this yet.

**Q: Is building dependencies from source too slow?**

A: So far, in this repository, no. But it does not rely on Ceres and OpenCV. And we won't want to rebuild on every run in CI. Vcpkg's support for [binary caching](https://learn.microsoft.com/en-us/vcpkg/users/binarycaching) is the next obvious optimization.

# TODO

- Define and (re-)implement package options to enable/disable testing, protos, etc.
- Finish refactor of farm_ng macros in `cmake/`
- Update and validate updated package `install` logic
- Specify desired third-party dependency versions in vcpkg.json
- Validate advanced packaging features: shared and static library support, etc.
- Add CI, including an integration test of a simple "consuming" project
- Add CMake linting if valuable


cpp/farm_ng/core/enum/CMakeLists.txt
```diff
@@ -21,15 +9,25 @@ farm_ng_add_library(farm_ng_core_enum
...
-                      LINK_LIBRARIES farm_ng_core_enum
-                      LABELS small)
+foreach(test_basename ${TEST_BASENAMES})
+  set(test_target ${test_basename}_test_farm_ng_core_enum)
+  set(test_src ${test_basename}_test.cpp)
+  add_executable(${test_target} ${test_src})
+  target_link_libraries(${test_target} farm_ng_core_enum GTest::gtest_main)
+  add_test(NAME ${test_target} COMMAND ${test_target})
+  set_property(TEST ${test_target} PROPERTY LABELS small)
 endforeach()
```
isherman, line 33: Changes in these CMakeLists.txt files are just experimenting with how farm_ng helper macros might be updated/refactored

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
+// For format details, see https://aka.ms/devcontainer.json. 
+// For config options, see https://github.com/devcontainers/templates/tree/main/src/ubuntu
+{
+  "name": "Ubuntu",
+  "image": "mcr.microsoft.com/devcontainers/base:jammy",
+  "customizations": {
```
wylderkeane, line 6: How is this going to work for arm builds? Will we use cross compilation and a toolchain or would we need to run this on an ARM host?

Also, what if I need a custom devcontainer for my project that requires additional volumes or other features?

.gitmodules
```diff
@@ -7,3 +7,12 @@
...
 	url = https://github.com/eigen-mirror/eigen.git
+[submodule "vcpkg"]
+	path = vcpkg
+	url = https://github.com/Microsoft/vcpkg.git
+[submodule "external/CLI11"]
+	path = external/CLI11
+	url = git@github.com:CLIUtils/CLI11.git
+[submodule "external/Eigen3"]
+	path = external/Eigen3
+	url = https://gitlab.com/libeigen/eigen.git
```
wylderkeane, line 18: Can we use `vcpkg` for these too?

CMakeLists.txt
```diff
@@ -1,18 +1,28 @@
-cmake_minimum_required(VERSION 3.16)
-# https://semver.org/
+cmake_minimum_required(VERSION 3.24)
```
wylderkeane, line 1: Can we avoid upgrading this? It's best practice to use the oldest version acceptable so building for older systems is still possible.

CMakeLists.txt
```diff
@@ -1,18 +1,28 @@
...
+
 project(farm_ng_core VERSION 0.2.0)
```
wylderkeane, line 3: Maybe we bump this to indicate all these changes? Some projects that build against this package would probably not be compatible out of the box anymore.

CMakeLists.txt
```diff
@@ -1,18 +1,28 @@
...
-set(CMAKE_CXX_CLANG_TIDY "")
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED True)
 
-find_package(fmt REQUIRED)
+enable_testing()
 
-add_subdirectory(protos)
 add_subdirectory(cpp)
+add_subdirectory(protos)
```
wylderkeane, line 28: I think it may be wise to generate protobuf files first so we can express dependencies on the protos from the cpp files. If I recall correctly, CMake won't recognize the dependency if we haven't generated it yet.

CMakePresets.json
```diff
@@ -0,0 +1,183 @@
...
+            "installDir": "${sourceDir}/out/install/${presetName}",
+            "cacheVariables": {
+                "BUILD_SOPHUS_TESTS": "Off",
+                "CMAKE_TOOLCHAIN_FILE": "${sourceDir}/vcpkg/scripts/buildsystems/vcpkg.cmake",
+                "CMAKE_CXX_FLAGS": "-Werror -Wall -Wextra -Wpedantic -Wno-unused-local-typedef -Wno-unused-local-typedef -Wno-gnu-zero-variadic-macro-arguments -Wno-unused-parameter -Wno-unused-variable -Wno-vla-extension",
+                "CMAKE_CXX_EXTENSIONS": "OFF",
+                "CMAKE_CXX_STANDARD": "20",
+                "CMAKE_CXX_STANDARD_REQUIRED": "ON",
+                "CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O2 -g",
+                "FARM_NG_PROVIDER_DEV_PACKAGES": "CLI11;Eigen3;Sophus"
```
wylderkeane, line 24: Why express this here instead of in the `CMakeLists.txt` file where it's invoked?

READMEv2.md
```diff
@@ -0,0 +1,110 @@
...
+# Install
+mkdir /tmp/farm-ng-prefix
+cmake --build --preset clang-debug --target install -DCMAKE_PREFIX_PATH=/tmp/farm-ng-prefix
+
+# Try the same workflows in Vscode
+# Select a configure preset such as "Clang RelWithDebInfo", then build and run as usual
+code farm_ng_core.code-workspace
+
+# Optionally, you can try the devcontainer workflow in Vscode by clicking "Reopen in Container" when prompted
+# - Note, this will not re-use the same vcpkg cache, so the first build will take more time
```
wylderkeane, line 88: Probably worth noting that if one must not reuse the same build folder inside and outside of the devcontainer, so it should be deleted entirely when switching between them.

cmake/farm_ng_module.cmake
```diff
@@ -5,57 +5,57 @@ include(CMakePackageConfigHelpers)
...
 macro(farm_ng_module)
 
-  message(STATUS "farm_ng_module ${PROJECT_NAME}")
-  # Configure CCache if available
-  find_program(CCACHE_FOUND ccache)
-  if(CCACHE_FOUND)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
-  endif(CCACHE_FOUND)
+  # message(STATUS "farm_ng_module ${PROJECT_NAME}")
```
wylderkeane, line 8: Unused now? How will this affect downstream users?

wylderkeane (commented):
Lots of good changes, but also lots of changes. How much of a ripple effect will this have to existing projects?

.gitmodules
isherman: For sure -- this was a demonstration of temporarily using development versions of these libraries. To switch back to the vcpkg-provided versions, you just remove the library name from the `FARM_NG_PROVIDER_DEV_PACKAGES` cache variable in `CMakePresets.json`

CMakeLists.txt
isherman: I've been watching a lot of cmake talks and one of the recommendations most talks make is not to be afraid of requiring a very recent cmake version. Cmake is still changing quickly, there are a lot of valuable APIs in recent versions, and the cmake binary itself is easy to upgrade (doesn't require sudo, or system-wide install, etc). 

But acknowledged, 3.24 _might_ not be required here.

CMakeLists.txt
isherman: Happy to do that. I'm not sure if this package has any downstream consumers that are consuming it as a package though.

CMakeLists.txt
isherman: Interesting. How would this issue manifest / is there a way to test that?

CMakePresets.json
isherman: The `CMakeLists.txt` job is just to express what dependencies the package depends on. It's the choice of an individual developer / packager where they want to get those dependencies from. For example, a developer might have a preset in their personal `CMakeUserPresets.json` that lets them switch between using the development and released versions of Sophus.

cmake/farm_ng_module.cmake
isherman: Not intending to merge like this, this PR is just a checkpoint of an experiment for discussion. 

One part of the experiment was refactoring these macros. So far in the experiment I think I've managed to factor out all the functionality of this macro into more appropriate places. But yes, if we were to merge like this it would break downstream repositories that rely on `farm_ng_module`.

.devcontainer/devcontainer.json
isherman: Good question, I haven't taken a shot at ARM builds, but I have a much better sense than before of how to express cross-compilation cleanly in CMake. Running on an ARM builder is also always an option. 

There is also the vcpkg support to consider. From a quick glance it looks like the public vcpkg repository does not include arm64-linux triplets in its CI (it does include arm64-windows). So to make vcpkg play well with arm64-linux we might need to use its "[overlay triplets](https://learn.microsoft.com/en-us/vcpkg/concepts/triplets)" feature?

We could definitely make derisking ARM builds a condition of pursuing this approach.

> Also, what if I need a custom devcontainer for my project that requires additional volumes or other features?

I think VsCode supports [specifying / switching between multiple devcontainers](https://github.com/microsoft/vscode-remote-release/issues/1165#issuecomment-1491325370). 

As you know you can specify the devcontainer as a Dockerfile; this is just using a higher-level API to show one example of a basic devcontainer.

isherman:
> Lots of good changes, but also lots of changes. How much of a ripple effect will this have to existing projects?

This PR is just a checkpoint of an experiment for discussion, it's not ready for merge and doesn't include an incremental migration plan. If we like (parts of) this, we can evaluate how/where/when we want to move in this direction.

isherman:
> * Move towards Sophus as separate project

This requires more consideration -- there is currently some cyclic dependeny between Sophus and farm-ng-core

strasdat:
For build on arm and other non-x86 targets one need to set

`export VCPKG_FORCE_SYSTEM_BINARIES=1`

before invoking

`./vcpkg/bootstrap-vcpkg.sh`

strasdat:
On Ubuntu 22.04, the latest cmake version is 22.x. Hence I used snap to get the latest cmake (28.x...)

`snap install cmake --classic`

strasdat:
I needed to add the following flags to the preset, but after this it build succesfully:

`-Wno-deprecated-anon-enum-enum-conversion -Wno-absolute-value`

This is on arm-linux with clang 14.

All (executed) test passed:

`ctest --preset clang-debug`

`100% tests passed, 0 tests failed out of 42`

This includes proto test (but not sophus tests).

PR #206, opened 2024-01-16
Fix pre-commit check / format.yml workflow

Hackerman342:


Hackerman342:
Found this issue that is exactly what we have going on: https://github.com/python/mypy/issues/16767

PR #209, opened 2024-02-13
feat: toShared() and copyToMut() for image

strasdat:


cpp/sophus/image/image_view.h
```diff
@@ -235,6 +235,17 @@ struct ImageView {
     return true;
   }
 
+  /// Returns an owned and shared Image. This often requires a copy of the data.
+  ///
+  /// The implementation is in image.h
+  Image<TPixel, Eigen::aligned_allocator<uint8_t>> toShared() const;
```
isherman, line 241: The API name `toShared()` -- is `shared` a term already used in the library? I wonder if `toImage` might be preferable, as an `Image` is already well-defined.

isherman (approved):

cpp/sophus/image/image_view.h
strasdat: Thanks for calling out this naming inconsistency.

The term "shared" is used throughout the image comments, and I believe it is important to highlight that an image with shared ownership is returned. Though I'm more leaning towards renaming Image to SharedImage in the long run.

[But I also like **Image** being a short name, though users are led to using it by default - which they should. And MutImage only if they must. ]

PR #210, opened 2024-02-14
Feat rig proto

strasdat:


cpp/farm_ng/core/proto_conv/sensor/conv.cpp
```diff
@@ -94,4 +94,26 @@ auto toProt<sophus::RigidCamera>(sophus::RigidCamera const& s)
   return proto;
 }
 
+template <>
+auto fromProt<core::proto::MultiCameraRig>(
+    core::proto::MultiCameraRig const& proto)
+    -> Expected<sophus::MultiCameraRig> {
+  sophus::MultiCameraRig cameras;
+  for (int i = 0; i < proto.cameras_size(); ++i) {
+    SOPHUS_TRY(auto, cam, fromProt(proto.cameras(i)));
```
isherman, line 103: should be FARM_TRY

isherman (approved):
Would be great to support `repeated` fields in a generic way in FARM_STRUCT in the future

cpp/farm_ng/core/proto_conv/sensor/conv.cpp
strasdat: oh, yes there are identical, but will fix it

PR #211, opened 2024-02-15
Add a read_payload() method to the events reader

wylderkeane:
To enable loading of raw payloads and interpreting them with the self-describing content.


isherman (approved):

edgarriba (commented):
Mind to extend with a unit test

PR #212, opened 2024-02-16
Feat message optional proto

strasdat:
Adding / refining the following four macros:

``FARM_CONV_IMPL_PRIMITIVE_OPTIONAL`` -  to/from proto conversion for std::optional of primitive types
``FARM_CONV_IMPL_MESSAGE_OPTIONAL`` - for non-primitive types
``FARM_CONV_IMPL_REPEATED_PRIMITIVE`` - to/from proto conversion for std::vector/std::deque of primitive types
``FARM_CONV_IMPL_REPEATED_MESSAGE`` - for non-primitive types

This PR also deletes the ``FARM_CONV_IMPL_OPTIONAL_ONE_ITER`` which did not add much value, but hampered code readability. 

Note:

The proto <-> c++ association is still to be 1:1, so one need to decided whether FARM_CONV_IMPL_REPEATED_MESSAGE shall target a std::vector or a std::deque.

TODO (future work):
Explore many to 1 proto <-> c++ conversion associations. This likely requires changes to the fromProt function templates (and underlying traits).

### DIFF
--- a/protos/farm_ng/core/std.proto
+++ b/protos/farm_ng/core/std.proto
@@ -22,6 +22,23 @@ package farm_ng.core.proto;
 // MesageNameG0ParameterName - e.g. OptionalG0Double which is
 // std::optional<double> in c++
 
+// Field naming pattern for primitives:
+//
+// ```
+// PRIMITIVE value = 1;
+// bool has_value = 2;
+// ````
+
+// Field naming pattern for messages:
+//
+// ```
+// MESSAGE value = 1;
+// bool has_message = 2;
+// ```
+//
+// ("has_value" is already auto-generaged for "value" in cpp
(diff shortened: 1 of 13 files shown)


protos/farm_ng/core/std.proto
```diff
@@ -22,6 +22,23 @@ package farm_ng.core.proto;
...
+// ````
+
+// Field naming pattern for messages:
+//
+// ```
+// MESSAGE value = 1;
+// bool has_message = 2;
+// ```
+//
+// ("has_value" is already auto-generaged for "value" in cpp
```
isherman, line 39: typo: "auto-generated"

isherman (approved):

PR #217, opened 2024-03-12
Support visit for a pair of images

isherman:


cpp/sophus2/image/image_types.h
```diff
@@ -132,6 +132,26 @@ struct ImageTraits<Eigen::Matrix<TT, kNumChannelsT, 1>> {
       std::is_floating_point_v<ChannelT> || std::is_unsigned_v<ChannelT>);
 };
 
+/// Visits the corresponding pixels of two ImageViews
+///
+/// Precondition: `lhs` and `rhs` are the same size and not empty
+template <class TPixel, class TFunc>
+auto visit(ImageView<TPixel> lhs, ImageView<TPixel> rhs, TFunc const& user_function) {
```
strasdat, line 139: Nit: the existing methods of a different name (though I do like visit better).

Ideal we would come up with a better naming scheme for all visitation methods.

strasdat (approved):

cpp/sophus2/image/image_types.h
isherman: OK, acknowledged. Probably not going to tackle that here.

PR #221, opened 2024-03-28
Add filename option to log formatting

wylderkeane:
The file option we use by default right now is very long and usually includes local paths on the host the software was built on. This modification provides a filename option to shorten the length of the prefix for the output while still retaining the ability to find where the log was coming from.

The default is updated to use the filename option as well.

### DIFF
--- a/cpp/farm_ng/core/logging/logger.cpp
+++ b/cpp/farm_ng/core/logging/logger.cpp
@@ -42,6 +42,9 @@ void StreamLogger::writeHeader(
     std::string const& file,
     int line,
     std::string const& function) {
+  auto filename_idx = file.rfind("/", std::string::npos);
+  auto parent_idx = file.rfind("/", filename_idx - 1);
+
   write(
       disk_logging_,
       FARM_FORMAT(
@@ -51,6 +54,8 @@ void StreamLogger::writeHeader(
           fmt::arg("text", header_text),
           fmt::arg("file", file),
           fmt::arg("line", line),
+          fmt::arg("fileparent", file.substr(parent_idx + 1)),
+          fmt::arg("filename", file.substr(filename_idx + 1)),
           fmt::arg("function", function)));
 }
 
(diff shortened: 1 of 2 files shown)


isherman (commented):
LGTM, I know @strasdat is opinionated on this front so I'll let him approve

cpp/farm_ng/core/logging/logger.cpp
```diff
@@ -51,6 +51,7 @@ void StreamLogger::writeHeader(
           fmt::arg("text", header_text),
           fmt::arg("file", file),
           fmt::arg("line", line),
+          fmt::arg("filename", std::filesystem::path(file).filename()),
```
strasdat, line 54: let's give it a single parent path for context as well

strasdat (changes requested):
Much better, let's just add the parent path.

cpp/farm_ng/core/logging/logger.cpp
wylderkeane: Done

cpp/farm_ng/core/logging/logger.cpp
```diff
@@ -51,6 +54,8 @@ void StreamLogger::writeHeader(
           fmt::arg("text", header_text),
           fmt::arg("file", file),
           fmt::arg("line", line),
+          fmt::arg("fileparent", file.substr(parent_idx + 1)),
```
strasdat, line 57: Huh, I never used it, but can't you just use: https://en.cppreference.com/w/cpp/filesystem/path/parent_path

strasdat (approved):

cpp/farm_ng/core/logging/logger.cpp
wylderkeane: That provides the entire parent path, not just the name of the parent directory.

isherman: I agree there's a cleaner way to do this with `std::filesystem` APIs than with `.substr`

PR #223, opened 2024-04-02
feat: batch/SIMD processiong feature for eigen/sophus2

strasdat:
**Core feature:** ``Batch<class TT, int kNum>`` template class, to enable full batch / SIMD processing

**How does this work?**

Eigen matrices are templates, where the first element is a Scalar type, e.g. ``double`` in this example: ``Eigen::Matirx3d = Eigen::Matrix<double, 3, 3>``.

However, there is a ``Eigen::NumTraits`` trait class, to support other "scalars". A relatively well-known example is ``ceres::Jet`` types. Here, ``ceres::Jet<...>`` is a dual number pair.

Similarly, we define a "scalar" type called ``Batch<...>``:

```
template <class TT, int kNum>
struct Batch {
  Batch() {}

 [...]

  Eigen::Array<TT, kNum, 1> b;
};
```

which contains an Eigen array ``b``.

And we specialize the ``Eigen::NumTraits`` template:

```
template <class TT, int kNum>
struct NumTraits<sophus2::Batch<TT, kNum>> {
  using Real = sophus2::Batch<TT, kNum>;
```

Now, we can perform batched matrix operations such as matrix x matrix multiplication:

```
  Eigen::Matrix<Batch<double, kNumBatches>, kM, kN> batched_mat_7x3;
  Eigen::Matrix<Batch<double, kNumBatches>, kN, kO> batched_mat_3x5;
  
  Eigen::Matrix<Batch<double, kNumBatches>, kM, kO> batched_mat_7x5 =
      batched_mat_7x3 * batched_mat_3x5;
```

Let's zoom in what is going on here. We note that kNumBatches is identical for all three matrices, and this is the critical aspect of these type of batch operations. Now, if kNumBatches is chosen wisely (e.g. kNumBatches==8), we will get SIMD (e.g. AVX on AMD64 CPUs) operations. 

The key aspect is, compared to using non-batch Eigen matrices, that we will get will get SIMD instructions, independent of what dimensions we choose for ``kM, kN, kO`` as long as `` kNumBatches`` is a multiple of 8 (or whatever the actual logic in Eigen is; the website is down at the moment) and corresponding compiler setttings are enabled.

_For AMD64 cpus,  you typically need to enable AVX instructions explicitly (https://stackoverflow.com/a/44962907). I did experimentally verify that the matrix matrix multiplcation in the test did genereate  AVX instructions - some month ago..._

### DIFF
--- a/cpp/sophus2/concepts/division_ring_prop_tests.h
+++ b/cpp/sophus2/concepts/division_ring_prop_tests.h
@@ -42,7 +42,7 @@ struct DivisionRingTestSuite {
           SOPHUS_ASSERT_WITHIN_REL(
               left_hugging.params(),
               right_hugging.params(),
-              10.0 * kEpsilonSqrt<Scalar>,
+              10.0 * kEpsilonSqrt<typename BatchTrait<Scalar>::Scalar>,
               "`(g1*g2)*g3 == g1*(g2*g3)` Test for {}, #{}/#{}/#{}",
               ring_name,
               params_id,
@@ -70,7 +70,7 @@ struct DivisionRingTestSuite {
           SOPHUS_ASSERT_WITHIN_REL(
               left_hugging.params(),
               right_hugging.params(),
-              kEpsilonSqrt<Scalar>,
+              kEpsilonSqrt<typename BatchTrait<Scalar>::Scalar>,
               "`g1 * g2 == g2 * g3` Test for {}, #{}/#{}",
               ring_name,
               params_id,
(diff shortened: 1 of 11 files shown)


cpp/sophus2/concepts/division_ring_prop_tests.h
```diff
@@ -91,16 +91,16 @@ struct DivisionRingTestSuite {
           Ring left_hugging = g1 * g2;
           Ring right_hugging = g2 * g1;
           ++num_cases;
-          if ((left_hugging.params() - right_hugging.params()).norm() <
-              kEpsilonSqrt<Scalar>) {
-            ++num_commutativity;
-          }
+          // if ((left_hugging.params() - right_hugging.params()).norm() <
```
strasdat, line 94: TODO: fix the tests in this file

cpp/sophus/linalg/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+#[[
```
strasdat, line 1: todo; remove stale file

PR #226, opened 2024-04-26
make StopWatch threadsafe

strasdat:


cpp/farm_ng/core/misc/stopwatch.h
```diff
@@ -36,33 +37,41 @@ class StopwatchSingleton {
...
-    std::chrono::duration<double> diff = stop_time - *b.maybe_start;
-    b.maybe_start.reset();
-    double d = diff.count();
-    if (d < b.min) {
-      b.min = d;
-    }
-    if (d > b.max) {
-      b.max = d;
+
+    if (b.maybe_start) {
```
isherman, line 60: I think this would be more readable with early-return style:
```
if (!b.maybe_start) {
  return 0.0;
}

...
```

cpp/farm_ng/core/misc/stopwatch.h
```diff
@@ -112,7 +123,7 @@ class StopwatchSingleton {
     double min = std::numeric_limits<double>::max();
     double max = std::numeric_limits<double>::lowest();
   };
-
+  std::mutex mutex_;
```
isherman, line 126: I like to comment what the mutex is protecting

isherman (approved):

PR #228, opened 2024-05-02
Preserve proto field names

guilhermedemouraa:


py/farm_ng/core/events_file_writer.py
```diff
@@ -43,7 +43,7 @@ def proto_to_json_file(file_path: str | Path, proto_message: Message) -> bool:
         return False
 
     with Path(file_path).open("w", encoding="utf-8") as file:
-        file.write(MessageToJson(proto_message))
+        file.write(MessageToJson(proto_message, preserving_proto_field_name=True))
```
Hackerman342, line 46: It is important not to break backwards compatibility for (1) public facing API's  & (2) utility functions like this that are used extensively in our codebase. In such cases, I recommend making the previous behavior of our function the default and allowing users to opt in to the new behavior we want in this one case.

```suggestion
def proto_to_json_file(file_path: str | Path, proto_message: Message, preserving_proto_field_name=False) -> bool:
    """Write a proto Message to a JSON file. The parent directory of the file must exist.
    Args:
        file_path (str | Path): The path to the JSON file to create / overwrite.
        proto_message (Message): The proto message to write to the JSON file.
        preserving_proto_field_name (bool): See MessageToJson in
                https://googleapis.dev/python/protobuf/latest/google/protobuf/json_format.html
      
    Returns:
        True if the file was written successfully, False otherwise.
    """
    
    .....
    
        file.write(MessageToJson(proto_message, preserving_proto_field_name=preserving_proto_field_name))
```

Hackerman342: @guilhermedemouraa I'll mark as resolved, but please take a look at the comment

guilhermedemouraa: > @guilhermedemouraa I'll mark as resolved, but please take a look at the comment

I did take a look at it!
Thanks

Hackerman342 (approved):

PR #238, opened 2024-06-04
Expose event_service.proto for farm_ng_amiga

ianBtesser:


guilhermedemouraa (approved):
This LGTM, but I'd double check with @edgarriba first.
Is there a reason why this wasn't exposed before?

edgarriba (approved):
LGTM. @guilhermedemouraa no reason

PR #250, opened 2024-07-24
Proto for live annotations

guilhermedemouraa:
Ticket: https://linear.app/farm-ng/issue/SWE-176/live-map-annotations

### DIFF
--- a/protos/farm_ng/core/recorder.proto
+++ b/protos/farm_ng/core/recorder.proto
@@ -37,7 +37,7 @@ message RecorderProfile {
 // A message that can be used to live annotate a recording.
 
 message RecorderAnnotation {
-    Timestamp stamp = 1; // The timestamp of the annotation.
+    repeated Timestamp stamp = 1; // The timestamp of the annotation.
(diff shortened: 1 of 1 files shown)


protos/farm_ng/core/recorder.proto
```diff
@@ -15,10 +15,29 @@
...
     string name = 1;
     repeated SubscribeRequest subscriptions = 2;
 }
+
+// A message that can be used to live annotate a recording.
+
+message RecorderAnnotation {
+    repeated farm_ng.core.proto.Timestamp stamps = 1; // The timestamp of the annotation.
+    AnnotationKind kind = 2; // The kind of annotation.
+    string message = 3; // An optional message of the annotation. E.g., "Start of row 1".
```
guilhermedemouraa, line 42: I think we should also add a field for geolocation. Either a `FilterState` or a `RelativePositionFrame.`

cc @mihelich, @lullabee, @Danvil

ianBtesser (approved):

PR #257, opened 2024-08-20
Input queue size

strasdat:


cpp/farm_ng/core/pipeline/input.h
```diff
@@ -89,18 +95,19 @@ class Input {
     // NOTE: seems that `io_context::strand::post` it's deprecated and need
     // to be use `boost::asio::post` instead. See:
     // https://www.boost.org/doc/libs/1_78_0/doc/html/boost_asio/reference/io_context__strand/post.html
-    context_strand_.getAsioStrand().post([this, value = std::move(value)] {
-      this->strandedSlot(std::move(value));
-    });
+    context_strand_.getAsioStrand().post(
+        [this, count, value = std::move(value)] {
+          this->strandedSlot(std::move(value), count);
```
ethanrublee, line 100: It seems to me that a more straightforward change would just be to expose a member function that returns the current atomic value of count_, ``int queueSize() const { return count_;}`` this function can then be called by the stranded_slot to determine if messages are backing up.

strasdat: They are both valuable pieces of information, maybe we should do both:

 - What was the queue length when this got posted? E.g. are we already behind when the input arrived.
 - What is the queue length NOW which can be called anytime from anywhere inside and outside the component.

PR #261, opened 2024-12-11
Lock req versions + release number

MScatolin:


setup.cfg
```diff
@@ -21,9 +21,12 @@ classifiers =
...
 [options]
 python_requires = >=3.8
+setup_requires =
+    wheel
 install_requires =
-    protobuf==5.27.2
-    grpcio==1.64.1
+    grpcio==1.56.2
+    grpcio-tools==1.56.2
+    protobuf==4.23.4
```
mihelich, line 29: @MScatolin this isn't what I expected. I thought we agreed on pinning `protobuf` to 5.27.x, and that was sufficient to resolve the problem for users of our SDK. In fact Kyle already made that change last month in #259, so I believe all that's left for `farm-ng-core` is to bump `version = 2.3.2` and do the PyPI release.

Yes, our internal image build in `workspace` currently builds these packages and our apps with `protobuf==4.23.4`. I don't believe we need to force such an old version on external users, who I assume were using 5.27.x successfully up until 5.28 was released.

mihelich (changes requested):
@MScatolin I thought were pinning `protobuf` to 5.27.x?

setup.cfg
MScatolin: My reasoning here is that, if we not using up-to-date protobuf and grpcio, better to stick to the versions in the Brain image to avoid possible future conflicts. LMK what the idea is and if you want me to change it.

mihelich: The other risk is package resolution conflicts for user applications. The rule of thumb for library packages is to make version requirements as lax as possible. `farm-ng-core` and `farm-ng-amiga` are library packages that users integrate into their applications. Those applications will have their own dependencies, some of which may also depend on `protobuf`. Constraining to a very old version like `protobuf==4.23.4` can wreak havoc on pip dependency resolution for user applications by creating outright conflicts or unexpectedly forcing other packages they depend on to old versions.

Suppose we require `protobuf==5.27.2` and some other dependency requires `protobuf==5.27.1`. Very likely there is no real conflict, and either version would work fine. But `pip install` would fail because it cannot satisfy both constraints.

We should actually say `protobuf<5.28`. Setting an upper bound is still bad form, but it makes sense when we know there's a conflict. I am not sure if the additional constraints on `grpcio` and `grpcio-tools` are actually necessary (in theory, if we constrain `protobuf`, `pip` should only resolve compatible versions of `grpcio` etc), but if so, they should also be upper bounds.

We should also have a comment here explaining that there's a conflict with 5.28 and referencing SWE-427.

MScatolin: Great. Thank you for the explanation. Updated to <=5.27.5 which is the latest version and tested by me and Ian.

ianBtesser (approved):

mihelich (approved):

setup.cfg
```diff
@@ -21,9 +21,11 @@ classifiers =
 
 [options]
 python_requires = >=3.8
+
 install_requires =
-    protobuf==5.27.2
-    grpcio==1.64.1
+    grpcio
+    protobuf<=5.27.5
+    farm-ng-package
```
mihelich, line 28:
```suggestion
```

`farm-ng-package` should be a build-time dependency only, can you remove it here?

setup.cfg
```diff
@@ -41,7 +43,8 @@ test =
     pylint-protobuf
     anyio
     types-protobuf
-    grpcio-tools==1.64.1
+    grpcio-tools
+    grpcio-tools
```
mihelich, line 47:
```suggestion
```

Remove duplicate `grpcio-tools` line.

mihelich (approved):
@MScatolin note clean-up comments, otherwise LGTM. Thanks!

setup.cfg
MScatolin: done

setup.cfg
MScatolin: done

PR #262, opened 2024-12-13
Fix Python CI breakages

mihelich:
Mitigate drift in supported GitHub runner versions:
1. `ubuntu-latest` now uses `ubuntu-24.04`, which broke our Python CI due to newer and stricter GCC.
2. `macos-12` is EOL.

Fix (1) by pinning Python builds to `ubuntu-22.04`. Note that `ubuntu-24.04` also drops out-of-the-box support for Python 3.8, which we should continue testing to match the Brain.

Fix (2) by upgrading to `macos-13`, likewise the final runner with Python 3.8 baked in. This required a fix to an undefined behavior (compiler-dependent) bug in the pybindings for Lie group types.

Completes SWE-531 and SWE-538

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,7 @@ jobs:
   # Based on https://dev.to/eugenebabichenko/automated-multi-platform-releases-with-github-actions-1abg
   create_release:
     name: Create Release and Provide Upload URL.
-    runs-on: ubuntu-22.04 # OS to create release, not for build
+    runs-on: ubuntu-20.04 # OS to create release, not for build
     outputs:
       # This job will provide URL for build jobs to use for uploading assets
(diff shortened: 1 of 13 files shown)


.github/workflows/release.yml
```diff
@@ -13,7 +13,7 @@ jobs:
   # Based on https://dev.to/eugenebabichenko/automated-multi-platform-releases-with-github-actions-1abg
   create_release:
     name: Create Release and Provide Upload URL.
-    runs-on: ubuntu-latest # OS to create release, not for build
+    runs-on: ubuntu-20.04 # OS to create release, not for build
```
guilhermedemouraa, line 16: Should this be `22.04`?

guilhermedemouraa (approved):
Very minor comment. Thanks!

PR #263, opened 2024-12-18
Hauke/swe 531 fix ci breakage

strasdat:


mihelich:
Thanks @strasdat! Closing as I've applied the fix in #262.

